PR #35, opened 2024-01-05
feat(cargo): consolidate versions and project organization

sasa-tomic:
- Use workspace versions, instead of per-project
- Rename the `release_cli` binary to `dre`


nikola-milosa (approved):
In my opinion its a nice change, probably should wait for more opinions.

pietrodimarco-dfinity (approved):

PR #42, opened 2024-01-05
Periodically clean bazel cache on the CI

sasa-tomic:
This pull request adds periodic cleaning of bazel cache.

It will run on every PR but will only delete the cache once it's older than 7 days.

We need this because bazel [does not clear old cache entries](https://github.com/bazelbuild/bazel/issues/1035), so the CI cache would try to grow indefinitely [until it crosses 10GB](https://github.com/actions/cache/blob/main/README.md#cache-limits), at which point it would stop uploading / refreshing completely.


bin/clean-bazel-cache.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
+#
+# Clean up everything in CACHE_DIR last modified more than DAYS days ago.
+#
+
+set -eEuo pipefail
+
+readonly CACHE_DIR="${1-${HOME}/.cache/bazel}"
+readonly DAYS="${2-7}"
```
LittleChimera, line 9: what does this mean?

LittleChimera (approved):

PR #44, opened 2024-01-05
chore(canisters): Move canisters from //rs to separate dir

sasa-tomic:
This should speed up builds, as we don't use any of the canisters yet.


canisters/node_status_canister/Cargo.toml
```diff
@@ -0,0 +1,7 @@
+[workspace]
+
+members = [
+    "src/node_status_canister_backend",
+]
+
+resolver = "2"
```
nikola-milosa, line 7: If we anticipate to have more canisters (which i think we do) shouldn't we put a workspace file in `canisters/Cargo.toml`?

sasa-tomic: Good point, let me move it

nikola-milosa (approved):
LGTM

PR #49, opened 2024-01-10
fix(ci): Use older Ubuntu image to use older GLIBC

sasa-tomic:
This should prevent issues such as 
image


LittleChimera (approved):

DFINITYManu:
Why not just put the binary in a newer container to run it?

PR #71, opened 2024-01-17
docs: Add docs for trustworthy node metrics

sasa-tomic:


nikola-milosa (approved):
LGTM! I would just ask for two additions in this PR that should've been covered earlier.
1. `.github/workflows/mkdocs.yaml` -> `on.push.paths` add one more path for `mkdocs.yml`
2. `mkdocs.yml` -> `repo_url` - should be https://github.com/dfinity/dre/
Thanks in advance

PR #75, opened 2024-01-19
fix(k8s): setting larger python image for cursor-initializator-elastic

nikola-milosa:


sasa-tomic (approved):
That could work, nice!

PR #88, opened 2024-01-22
Improve XDR explanation

yvonneanne:


nikola-milosa (approved):
Looks good to me, but I would ask @sasa-tomic for another set of eyes

sasa-tomic (approved):
LGTM, thanks @yvonneanne !

PR #91, opened 2024-01-22
Scrape GuestOS metrics-proxy and clean up issues in multiservice-discovery.

DFINITYManu:
`multiservice-discovery` was incorrectly discovering HostOS node exporter as GuestOS node exporter.

The matter has been corrected.  This is a sample diff from its prior output and the current output.  Note IP address of `host_node_exporter` in the prior scrape.

image


nikola-milosa:
Close this and rename the branch to `container-...` so we can get the images built. Note that in order for this to work you need to update all service discovery and downloaders in k8s repo

PR #93, opened 2024-01-22
Scrape GuestOS metrics-proxy and clean up issues in multiservice-discovery. #91

DFINITYManu:
`multiservice-discovery` was incorrectly discovering HostOS node exporter as GuestOS node exporter.

The matter has been corrected.  This is a sample diff from its prior output and the current output.  Note IP address of `host_node_exporter` in the prior scrape.

image


nikola-milosa (approved):
Great job, thank you a lot! If you need any help in deploying ping me

pietrodimarco-dfinity (approved):

PR #96, opened 2024-01-23
fix(dashboard): Automatically update the internal dashboard k8s deployment

sasa-tomic:


nikola-milosa (approved):
Looks pretty cool, I think there is a github action for pushing to gitlab so maybe if we did this as two step process in the future we could reduce the size of this code

.github/workflows/bazel.yaml
```diff
@@ -62,6 +62,40 @@ jobs:
...
+        if: ${{ startsWith(github.ref, 'refs/tags/v') || startsWith(github.ref, 'refs/heads/container') || (github.ref == 'refs/heads/main') }}
         run:
           bazel query --noshow_progress 'kind("oci_push", ...)' | xargs -I_target bazel run _target -- --tag ${GITHUB_SHA}
+      - name: Update k8s deployments
+        if: ${{ startsWith(github.ref, 'refs/tags/v') || startsWith(github.ref, 'refs/heads/container') || (github.ref == 'refs/heads/main') }}
+        run: |
+            set -eExou pipefail
+
+            # checkout branch
+            apk add git
```
LittleChimera, line 74:
```suggestion
```

LittleChimera, line 74:
```suggestion
```

PR #98, opened 2024-01-23
fix(ci): Update the reference to the gitlab api token

sasa-tomic:


.github/workflows/bazel.yaml
```diff
@@ -76,7 +76,7 @@ jobs:
             set -eExou pipefail
 
             # checkout branch
-            git clone "https://gitlab-ci-token:${GITLAB_API_TOKEN}@gitlab.com/dfinity-lab/private/k8s/k8s.git"
+            git clone "https://gitlab-ci-token:${{ secrets.GITLAB_API_TOKEN }}@gitlab.com/dfinity-lab/private/k8s/k8s.git"
```
NikolaMilosa, line 79: I think this won't work. You have to set it as a global env var like [here](https://docs.github.com/en/actions/learn-github-actions/variables) and then reference it as environment variable. I don't think they expand it there

nikola-milosa (approved):

PR #100, opened 2024-01-23
Add test for multiservice-discovery

pietrodimarco-dfinity:
Runs multiservice-discovery with local registry created from `large02` testnet and execute tests on `/prom/targets` endpoint.
Add Bazel test:
`//rs/ic-observability/multiservice-discovery:prom_targets_tests          PASSED in 5.0s`


nikola-milosa:
I think this should be synced with #93?

pietrodimarco-dfinity:
> I think this should be synced with #93?
We don't have any metrics_proxy in testnet currently we can modify the test once it is merged

NikolaMilosa (approved):
Thanks!

PR #103, opened 2024-01-23
fix(ci): Cloning k8s repo to a suitable directory

sasa-tomic:
Should fix https://github.com/dfinity/dre/actions/runs/7626415516/job/20772752035
> Run set -eExou pipefail
> + git clone ***gitlab.com/dfinity-lab/private/k8s/k8s.git
fatal: destination path 'k8s' already exists and is not an empty directory.
Error: Process completed with exit code 128.

### DIFF
--- a/.github/workflows/bazel.yaml
+++ b/.github/workflows/bazel.yaml
@@ -84,6 +84,8 @@ jobs:
         run: |
             set -eExou pipefail
 
+            cd .git
+
             # checkout branch
             git clone "https://gitlab-ci-token:${GITLAB_API_TOKEN}@gitlab.com/dfinity-lab/private/k8s/k8s.git"
 


nikola-milosa (approved):
Generally bazel.yaml starts to have a lot of things in it. Maybe we should dispatch another workflow from thisone via workflow call and have this logic in another workflow since it doesn't require bazel or this repo, just the commit, what do you think?

sasa-tomic:
> Generally bazel.yaml starts to have a lot of things in it. Maybe we should dispatch another workflow from thisone via workflow call and have this logic in another workflow since it doesn't require bazel or this repo, just the commit, what do you think?

I agree. However, the one thing blocking us from doing that is the dependency between bazel and these other jobs. We'll have to be careful to not copy code around multiple times.

PR #107, opened 2024-01-24
Fix multiservice-discovery to use mainnet registry

pietrodimarco-dfinity:


rs/archive-utils/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+reqwest = { workspace = true , features = ["stream"]  }
+flate2 = { workspace = true }
+tokio = { workspace = true }
+anyhow = { workspace = true }
+tar = { workspace = true }
+futures-lite = { workspace = true}
+flume = "0.11.0"
```
sasa-tomic, line 15: We already use crossbeam, is flume better?

rs/archive-utils/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "archive-utils"
+version = { workspace = true }
+edition = { workspace = true }
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+reqwest = { workspace = true , features = ["stream"]  }
```
sasa-tomic, line 9: no need to re-specify the feature here

PR #113, opened 2024-01-25
ci: attempting to cleanup ci

NikolaMilosa:
IMHO `bazel.yaml` has become really large and here is an attempt to make it better. Looking forward for your opinions!


.github/workflows/bazel.yaml
```diff
@@ -1,73 +1,17 @@
-name: Bazel
+name: Build and test
```
sasa-tomic, line 1: I like the name change

.github/workflows/update-dependancies.yaml
```diff
@@ -0,0 +1,43 @@
...
+      - uses: ./.github/workflows/setup-bazel
+
+      ########################################
+      # Once per night, update dependencies and completely delete and recreate bazel cache
+      ########################################
+      - uses: actions/setup-python@v4
+        with:
+          python-version: "3.11"
+      - uses: snok/install-poetry@v1
+      - name: If running on schedule (at night) and on main, completely delete bazel cache then update deps
```
sasa-tomic, line 20: We also need to run bazel build in this schedule, otherwise the bazel cache will not contain bazel dir and the first build in the day will take very long time.

.github/workflows/bazel.yaml
```diff
@@ -127,41 +61,7 @@ jobs:
...
-            fi
-
-            # Push changes and create a new merge request
-            git commit -m "Updating container base image refs"
-            git push \
-                -o merge_request.create \
-                -o merge_request.title="[nomrbot] - Updating container image refs mainnet-dashboard [$GITHUB_SHA]" \
-                -o merge_request.description="Changes to the release repository - [here](https://github.com/dfinity/dre/commit/$GITHUB_SHA)" \
-                --force --set-upstream origin "update-image-tag-${GITHUB_SHA}"
+        uses: ./.github/workflows/update-deployments.yaml
```
sasa-tomic, line 64: Is the plan to make this more generic than k8s deployments? If not, would be better to be more specific (include k8s in the file name).

.github/workflows/prepare-release.yaml
```diff
@@ -0,0 +1,38 @@
...
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          body_path: CHANGELOG.md
+          generate_release_notes: true
+          draft: true
+          prerelease: true
+          files: |
+            release/*
+
+      - uses: geekyeggo/delete-artifact@v4
```
sasa-tomic, line 36: What does this do?

.github/workflows/update-dependancies.yaml
NikolaMilosa: Yup, my bad, fixed

.github/workflows/bazel.yaml
NikolaMilosa: This is super specific, I think this cannot be made more generic for now. Will rename

.github/workflows/prepare-release.yaml
NikolaMilosa: Wanted to abstract creating a release here so if we ever go to a scheduled release mode rather than manual, we can use this code for that. In `bazel.yaml` we upload build binary to the artifact storage of the repository and when we call this workflow we want to remove it since it is added in the release.

.github/workflows/main.yaml
```diff
@@ -0,0 +1,70 @@
+name: Build and test
+on:
+  push:
+    branches:
+      - "main"
+  pull_request:
+  merge_group:
+jobs:
+  bazel:
+    runs-on: ubuntu-22.04
```
sasa-tomic, line 10:
```suggestion
    runs-on: ubuntu-20.04
```

NikolaMilosa: Added

.github/workflows/update-dependancies.yaml
```diff
@@ -0,0 +1,48 @@
...
+          #
+          # Update dependencies
+          #
+          cargo update
+          poetry update
+          ./bin/poetry-export.sh
+          CARGO_BAZEL_REPIN=true bazel query //...
+
+      - name: Build
+        run: bazel buld ...
```
sasa-tomic, line 37:
```suggestion
        run: bazel build ...
```

.github/workflows/prepare-release.yaml
sasa-tomic: Understood, thanks!

PR #116, opened 2024-01-25
Simplify the definition management code in multiservice-discovery.

DFINITYManu:
I did not get to do this earlier because we had crunch, but now that I have a little bit more time, I have invested it into cleaning up this code to remove duplication and make the two management operations more symmetrical.

There is another cleanup I have in mind which is to marry the thread join handle into the definition so that we don't have to carry two separate `Vec[]`s (one for definitions and one for their join handles), which is nasty and can induce bugs later.  That will have to wait until at least tomorrow, though.

Functionality is equivalent and has not changed.


NikolaMilosa (approved):
Since we are in a hurry I will come back to this code tomorrow, but on a quick look seems okay.

PR #120, opened 2024-01-26
fix(ci): Correctly pass secrets to the k8s-deployments composite actions

sasa-tomic:
Composite actions do not have access to secrets, by default. Secrets need to be explicitly passed to them.


NikolaMilosa (approved):
Thanks for the fixes

PR #127, opened 2024-01-29
ci: fixing leftover name of step

NikolaMilosa:
Fixing leftover name of refactoring from here:
https://github.com/dfinity/dre/actions/runs/7690062449/job/20953364937


sasa-tomic (approved):
Thanks

PR #132, opened 2024-01-29
Fix test in service-discovery to be applied in mainnet registry

pietrodimarco-dfinity:
* Add mainnet test using Bazel


NikolaMilosa (approved):
Its a great start, thanks Pietro!

PR #146, opened 2024-02-01
feat(dre): added registry dump and migrated dre to use `slog`

NikolaMilosa:
The PR allows you to:
1. Do a dump of the mainnet registry of latest version. Here a temporary folder will be created and deleted at the end of the sync.
```bash
bazel run //rs/cli:dre -- dump-registry
```
2. Do a dump of a custom networks registry of some version.
```bash
bazel run //rs/cli:dre -- --network <nns-url> dump-registry --version <version> 
```
3. Do a dump of a cached registry. The cached registry will first be upgraded to the latest version and then the specified version will be dumped. If the `--path` parameter is passed the cached registry will not be destroyed at the end of the action.
```bash
bazel run //rs/cli:dre -- dump-registry --version <version> --path <path>
```
The dre tool was migrated to use `slog` as a logging solution, since our other binaries use it as well and the crate `service-discovery` which got integrated here depends on it.


rs/cli/src/main.rs
```diff
@@ -379,17 +385,9 @@ fn local_unused_port() -> u16 {
...
-            if std::env::var("LOG_LEVEL").is_err() {
-                // Default logging level is: info generally, warn for mio and actix_server
-                // You can override defaults by setting environment variables
-                // RUST_LOG or LOG_LEVEL
-                std::env::set_var("LOG_LEVEL", "info,mio::=warn,actix_server::=warn")
-            }
-        }
-    }
-    pretty_env_logger::init_custom_env("LOG_LEVEL");
+fn make_logger() -> Logger {
```
sasa-tomic, line 388: How do we set / keep the log level filtering functionality?

rs/cli/src/cli.rs
```diff
@@ -1,8 +1,10 @@
+use std::path::PathBuf;
+
 use clap::{Parser, Subcommand};
 use clap_num::maybe_hex;
 use ic_base_types::PrincipalId;
 use ic_management_types::{Artifact, Network};
-use log::error;
+use slog::{error, Logger};
```
sasa-tomic, line 7: Ideally, switch from log -> slog would have gone into a separate PR. Having it together with the other change makes it harder to review.

rs/cli/src/cli.rs
```diff
@@ -120,6 +122,16 @@ pub(crate) enum Commands {
         /// all subnets
         subnet_ids: Vec<PrincipalId>,
     },
+    /// Dump full registry as json
+    DumpRegistry {
```
sasa-tomic, line 126: We should think in the future about putting all registry-related ops into a separate subcommand. I don't see an immediate need for this since we have only one function for the registry, which is "dump". But we can revisit that later.

rs/cli/src/registry_dump.rs
```diff
@@ -0,0 +1,161 @@
...
+            let binding = format!("~/tmp/{}", uuid);
+            let local_temp = shellexpand::tilde(&binding);
+            (
+                PathBuf::from_str(&local_temp).map_err(|e| anyhow::anyhow!("Error converting to path: {:?}", e))?,
+                true,
+            )
+        }
+    };
+
+    sync_local_registry(logger.clone(), path.clone(), vec![network.get_url()], false, None, None)
```
sasa-tomic, line 39: We already have a similar function in the dashboard backend, and it's accessible from the cli as well. Is there a good reason why we need to have both?

rs/cli/src/registry_dump.rs
```diff
@@ -0,0 +1,161 @@
...
+        .get_family_entries_versioned::<DataCenterRecord>(version)
+        .map_err(|e| anyhow::anyhow!("Couldn't get data centers: {:?}", e))?;
+    let dcs = serde_json::to_string(&dcs).map_err(|e| anyhow::anyhow!("Couldn't convert to JSON: {:?}", e))?;
+
+    let node_operators = client
+        .get_family_entries_versioned::<NodeOperatorRecord>(version)
+        .map_err(|e| anyhow::anyhow!("Couldn't get data centers: {:?}", e))?;
+    let node_operators =
+        serde_json::to_string(&node_operators).map_err(|e| anyhow::anyhow!("Couldn't convert to JSON: {:?}", e))?;
+
```
sasa-tomic, line 77: There is also node providers and node rewards that come to my mind.

sasa-tomic (approved):

rs/cli/src/cli.rs
NikolaMilosa: The integrated functionality requires serilog so I had to add it all in one PR

rs/cli/src/registry_dump.rs
NikolaMilosa: I didn't see that backand was a lib as well. I will integrate that! Thanks!

rs/cli/src/registry_dump.rs
NikolaMilosa: I will close this PR since if we integrate the backend code we don't need serilog yet. So to not complicate it I will migrate

PR #148, opened 2024-02-01
Upload test artifacts from main

pietrodimarco-dfinity:
Then we would need to use https://github.com/marketplace/actions/download-workflow-artifact to fetch the artifacts from a different workflow as https://docs.github.com/en/actions/managing-workflow-runs/downloading-workflow-artifacts doesn't allow this


sasa-tomic (approved):
Brilliant!

PR #150, opened 2024-02-02
feat(dre): added registry dump

NikolaMilosa:
The PR allows you to:
1. Do a dump of the mainnet registry of latest version. Here a temporary folder will be created and deleted at the end of the sync.
```bash
bazel run //rs/cli:dre -- dump-registry
```
2. Do a dump of a custom networks registry of some version.
```bash
bazel run //rs/cli:dre -- --network <nns-url> dump-registry --version <version> 
```
3. Do a dump of a cached registry. The cached registry will first be upgraded to the latest version and then the specified version will be dumped. If the `--path` parameter is passed the cached registry will not be destroyed at the end of the action.
```bash
bazel run //rs/cli:dre -- dump-registry --version <version> --path <path>
```


pietrodimarco-dfinity (approved):
Very nice :) THX

PR #151, opened 2024-02-02
feat(dre): implementing a protection method for updating unassigned nodes

NikolaMilosa:
To test out run:
```bash
cargo run --bin dre -- update-unassigned-nodes --nns-subnet-id <subnet_id>
```
The cli fetches the version of the nns of the specified network (mainnet by default). If the target is not on mainnet then the user should provide a specific `nns-subnet-id` so the tool knows which subnet to use to compare the versions.


sasa-tomic (approved):
Very nice, thanks!

PR #152, opened 2024-02-02
Fixed sync_and_stop to load in mem cache

pietrodimarco-dfinity:
The following provides a wrapper around `RunningDefinition` used for testing.
It fixes previous `sync` to also load the in-mem cache


NikolaMilosa (approved):
Lets see if this helps! Thanks

PR #153, opened 2024-02-02
Fix multiservice discovery test

pietrodimarco-dfinity:


rs/ic-observability/multiservice-discovery/README.md
```diff
@@ -12,6 +12,31 @@ networks on the fly.
...
 
+## Test
+
+We are performing these checks to ensure that the multiservice-discovery fetches all the necessary targets
+before deploying it in production.
+If less or erroneus targets end up in production, we risk compromising the observability stack leading
+to incomplete or erroneus observability data.
+If the mainnet registry gets updated or if the logic for target discovery is modified,
+this assertions should be updated accordingly.
+One can run the multiservice-discovery from local and manually update the targets after inspection:
```
sshtmc, line 23:
```suggestion
Integration tests check if the multiservice-discovery lists all expected targets.
If less than expected, or if some incorrect targets end up in production, we risk compromising the entire observability stack and the public dashboard.
Whenever the mainnet registry gets updated or if the logic for target discovery is modified, the tests need to be updated accordingly.

The tests can be run locally for inspection:

```

rs/ic-observability/multiservice-discovery/tests/tests.rs
```diff
@@ -0,0 +1,144 @@
...
+                        } else {
+                            let mut new_set = BTreeSet::new();
+                            new_set.insert(value);
+                            acc.insert(key, new_set);
+                        }
+                    }
+                    acc
+                });
+
+        assert_eq!(targets.len(), 7374);
```
sshtmc, line 75: This will be painful to update all the time, and will lead to unpredictably and uncontrollably failing tests. We'll have to find a better way to do this. For instance, add an error margin, and periodically and automatically update the expected value, based on the reference value provided by the code from "main". But this can come in follow-up PRs, I believe.

bazel-dre-1
```diff
@@ -0,0 +1 @@
+/private/var/tmp/_bazel_pietro.di.marco/b166ba5137d2b5b80ccec0055091def4/execroot/dre
```
NikolaMilosa, line 1: Is this needed or is it a leftover?

rs/ic-observability/multiservice-discovery/README.md
```diff
@@ -12,6 +12,31 @@ networks on the fly.
...
+
+# Check number of targets:
+jq -r '[.[] | .targets[]] | length' targets.json
+
+# List unique jobs:
+jq -r '.[].labels.job' targets.json | sort -u
+
+# List unique subnets:
+jq -r '.[].labels.ic_subnet'  targets.json | sort -u
+```
```
NikolaMilosa, line 38: Is there a particular reason why we use prom format here? Feels like our format for displaying targets is more natural and a must-have feature while this is just a convenience that we added.

rs/ic-observability/multiservice-discovery/tests/tests.rs
```diff
@@ -0,0 +1,144 @@
...
+                    let targets: Vec<PrometheusStaticConfig> = serde_json::from_str(&text).unwrap();
+                    return Ok(targets);
+                }
+            }
+            thread::sleep(Duration::from_secs(5));
+        }
+    }
+
+    #[test]
+    fn prom_targets_tests() {
```
NikolaMilosa, line 38: For asserting things here I suggest that we use the public dashboard API. Here is the [swagger](https://ic-api.internetcomputer.org/api/v3/swagger) of their API and we could fetch all the information from there and check what we need. For eg to get all the nodes just use 
```bash
curl -X 'GET' \
  'https://ic-api.internetcomputer.org/api/v3/nodes?include_status=DOWN&include_status=UP&include_status=UNASSIGNED&include_status=DEGRADED' \
  -H 'accept: application/json'
```
You have an equivalent for subnets as well. Probably easier than to periodically update this test

NikolaMilosa (commented):
Looks okay, the asserting with public API would be spot on IMHO

rs/ic-observability/multiservice-discovery/tests/tests.rs
pietrodimarco-dfinity: Makes sense @NikolaMilosa ;) thx!

rs/ic-observability/multiservice-discovery/README.md
pietrodimarco-dfinity: Probably not needed anymore when we switch to public dashboard API

bazel-dre-1
pietrodimarco-dfinity: eah that's a leftover let me get rid of it

rs/ic-observability/multiservice-discovery/tests/tests.rs
```diff
@@ -0,0 +1,131 @@
...
+            }
+            thread::sleep(Duration::from_secs(5));
+        }
+    }
+
+
+    async fn expected_nodes_and_subnets() -> Result<(Vec<String>, Vec<String>), reqwest::Error> {
+        const API_NODES_URL: &str = "https://ic-api.internetcomputer.org/api/v3/nodes";
+        let response = reqwest::get(API_NODES_URL).await?.text().await?;
+        let response_value: Value = serde_json::from_str(&response).unwrap();
```
NikolaMilosa, line 43: You should be able to do to `let response: Value = reqwest::get(API_NODES_URL).await?.json().await?;` in one line.

rs/ic-observability/multiservice-discovery/tests/tests.rs
```diff
@@ -0,0 +1,131 @@
...
+                    subnets.insert(String::from(sub));
+                }
+
+                String::from(val["node_id"].as_str().unwrap())
+            }).collect::<Vec<_>>();
+            
+        Ok((nodes, subnets.into_iter().collect()))
+    }
+
+    #[test]
```
NikolaMilosa, line 59: I've seen that there is [`tokio::test`](https://docs.rs/tokio/latest/tokio/attr.test.html) where you can skip all these runtime things. Haven't tried it tho

rs/ic-observability/multiservice-discovery/tests/tests.rs
```diff
@@ -0,0 +1,131 @@
...
+    fn prom_targets_tests() {
+        let rt = Runtime::new().unwrap();
+
+        let (expected_nodes, expected_subnets) = rt.block_on(rt.spawn(async { 
+            expected_nodes_and_subnets().await 
+        })).unwrap().unwrap();
```
NikolaMilosa, line 65:
```suggestion
        let (expected_nodes, expected_subnets) = rt.block_on(expected_nodes_and_subnets()).unwrap();
```

rs/ic-observability/multiservice-discovery/tests/tests.rs
```diff
@@ -0,0 +1,131 @@
...
+            registry_dir.path().to_str().unwrap(),
+        ];
+        let mut sd_server = Command::cargo_bin(CARGO_SD_BIN)
+            .unwrap_or(Command::new(BAZEL_SD_BIN))
+            .args(args)
+            .spawn()
+            .unwrap();
+        let targets = rt.block_on(rt.spawn(async { 
+            fetch_targets().await 
+        })).unwrap().unwrap();
```
NikolaMilosa, line 80:
```suggestion
        let targets = rt.block_on(fetch_targets()).unwrap();
```

NikolaMilosa (approved):
Looks great! Some small things I would improve 🚀  but aren't pressing

rs/ic-observability/multiservice-discovery/tests/tests.rs
```diff
@@ -0,0 +1,131 @@
...
+                    return Ok(targets);
+                }
+            }
+            thread::sleep(Duration::from_secs(5));
+        }
+    }
+
+
+    async fn expected_nodes_and_subnets() -> Result<(Vec<String>, Vec<String>), reqwest::Error> {
+        const API_NODES_URL: &str = "https://ic-api.internetcomputer.org/api/v3/nodes";
```
sasa-tomic, line 41: This is a very cool idea. Nice!
However, we should also think of a way to handle non-working public dashboard API, since we now introduce hard dependence on them being online at all times for our tests to work. And they don't have 100% uptime. And if they are down, we may not be able to (quickly) merge a PR that's needed to bring them back up.

sasa-tomic (approved):
Beautiful, thanks!

rs/ic-observability/multiservice-discovery/tests/tests.rs
NikolaMilosa: The only other thing I see is to spin up a whole network in the CI and use that. Generally we will never be able to have 100% non-flaky test

sasa-tomic: I was thinking more of making the process of downloading the contents more robust. I'll add more comments below.

rs/ic-observability/multiservice-discovery/tests/tests.rs
```diff
@@ -0,0 +1,131 @@
...
+        let response = reqwest::get(API_NODES_URL).await?.text().await?;
```
sasa-tomic, line 42: We should add a [timeout](https://docs.rs/reqwest/latest/reqwest/struct.RequestBuilder.html#method.timeout) and retries here, since the API may be partially or fully unavailable.

rs/ic-observability/multiservice-discovery/tests/tests.rs
pietrodimarco-dfinity: @sasa-tomic I want to add support for create the `expected_data`  by running the SD from the main bin. however we need a way to publishing the artifact public.

NikolaMilosa (approved):
You're da b-o-s-s thanks!

PR #164, opened 2024-02-05
feat(ci): Bump up the actions/cache to v4, and force overwriting the cache

sasa-tomic:
The force overwriting implementation comes from https://github.com/actions/cache/blob/a2ed59d39b352305bdd2f628719a53b2cc4f9613/tips-and-workarounds.md?plain=1#L38


NikolaMilosa (approved):
Great! Thanks

PR #166, opened 2024-02-06
feat(k8s): adding parameter for ic name in boundary node service discovery

NikolaMilosa:
Forgotten fix when the `ic` parameter for `mainnet` in `multiservice-discovery` was moved to `mercury`


sasa-tomic (approved):
Thanks!

PR #169, opened 2024-02-07
Load and Store definitions from/to disk

pietrodimarco-dfinity:
The following add definitions persistence to MSD. 
https://dfinity.atlassian.net/browse/OB-496.


NikolaMilosa (approved):
Looks great, I think having it as "opt-in" is the best for now. Thanks!

PR #173, opened 2024-02-08
Fix release notes generator.

DFINITYManu:


NikolaMilosa (approved):
Nice catch! Thanks

sasa-tomic (approved):
Great catch!

PR #174, opened 2024-02-08
Repin for Bazel if needed in Build step

pietrodimarco-dfinity:
* Added composite step to build and run bazel REPIN if needed
* The action pushes back [Cargo.Bazel.lock] in the branch


.github/workflows/main.yaml
```diff
@@ -24,7 +24,17 @@ jobs:
...
       ########################################
       - name: "🚀 Building"
-        run: bazel build ...
+        run: |
+          set +e -x
+          build_output=$(bazel build ... 2>&1)
+          exit_code=$?
+          if [[ $exit_code -ne 0 ]] && grep -q "Digests do not match:" <<< "$build_output"; then
+              export CARGO_BAZEL_REPIN=true
+              build_output=$(bazel build ... 2>&1)
```
sasa-tomic, line 33: The changed lock files still need to be pushed back to the branch, or shall we do that in a follow up?

.github/workflows/main.yaml
```diff
@@ -24,7 +24,17 @@ jobs:
       # Build and test
       ########################################
       - name: "🚀 Building"
-        run: bazel build ...
+        run: |
+          set +e -x
+          build_output=$(bazel build ... 2>&1)
```
DFINITYManu, line 29: You must actually output this, not just swallow it.

Here is a way you can do this properly:

```
MYTMPDIR="$(mktemp -d)"
trap 'rm -rf -- "$MYTMPDIR"' EXIT
r=0
logfile="$MYTMPDIR/bazel-out.log"
bazel build ... 2>&1 | tee "$logfile" >&2 || r=?
if [ "$r" != 0 ]; then
    if grep -q 'Digests do not match" "$logfile" ; then
        export CARGO_BAZEL_REPIN=true
        bazel build ...
    else
        exit $r
    fi
fi
````

**Review by DFINITYManu — CHANGES_REQUESTED** [review]

File: .github/workflows/main.yaml (hunk 2 above)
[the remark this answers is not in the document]
Comment (pietrodimarco-dfinity) [reply .github/workflows/main.yaml:29]: Thx :)

File: .github/workflows/build/action.yaml [hunk 3]
```diff
@@ -0,0 +1,54 @@
...
+                if grep -q 'Digests do not match' "$logfile" ; then
+                    echo "repin=true" >> $GITHUB_OUTPUT
+                    return 0
+                else
+                    return $r
+                fi
+            fi
+        }
+        
+        bazel_build_auto_repin build ...
```
NikolaMilosa, line 32: I think its a bit lighter to use `query` for repinning, can we not use that?

sasa-tomic (approved):

NikolaMilosa (approved):
Noice!

PR #175, opened 2024-02-09
feat(registry-dump): Interpret vec<u8> data in registry structs as corresponding types

sasa-tomic:
- Instead of expecting from the user to interpret Vec<u8> as principals (somehow), we can do this in the code directly, and show principals as principals
- Make the dump more in line with what ic-admin shows (in ic-admin get-topology)
- Show the node reward table as well
- Use native JSON serialize for the dump, instead of manually crafting the wrapper struct
- Use the same dir that the registry sync uses, to avoid the slow re-downloading the entire mainnet registry on every execution


NikolaMilosa (approved):
Great thanks!

pietrodimarco-dfinity (approved):

PR #187, opened 2024-02-12
Refactor metrics to simplify updating the values

DFINITYManu:
Use a private data structure to track per-network metrics, then use that data structure as data source for the metrics callbacks.  The once-added callbacks can then serve the metrics for registered definitions.

In addition to that, when a definition ends, we deregister it from the per- network metrics data structure, so we stop serving metrics about it.

We are no longer using Tokio mutexes since this would require async closures for the callbacks, and async closures are unstable.  std::sync::Mutex works perfectly OK for this use and, in fact, Tokio documentation states that standard mutexes can work fine with Tokio code, and should be preferred in most cases.

https://docs.rs/tokio/1.24.2/tokio/sync/struct.Mutex.html#which-kind-of-mutex-should-you-use indicates we should use the standard mutex when we are not going to hold the lock across await points.  Since that is exactly what we are doing (there are no awaits involved in the set() and get() code for the data structure being locked), then we can simply use the simple blocking mutex, which is what we have done here.

### DIFF
--- a/rs/ic-observability/multiservice-discovery/src/definition.rs
+++ b/rs/ic-observability/multiservice-discovery/src/definition.rs
@@ -488,7 +488,6 @@ impl DefinitionsSupervisor {
                     let fs_def: Vec<FSDefinition> = existing
                         .values()
                         .cloned()
-                        .into_iter()
                         .map(|running_def| running_def.definition.into())
                         .collect::<Vec<_>>();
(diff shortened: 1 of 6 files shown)


sasa-tomic (approved):
Looks much better, nice cleanup. Thanks Manuel!

PR #188, opened 2024-02-13
chore(msd): Improvements and cleanup

NikolaMilosa:
In this PR we have:
* removed some leftover `&mut self` that were not needed
* moved from `Mutex` to `RwLock` since when sending the metrics we don't need other write threads to stop which improves perf
* fixed the issue where we return `Ok(())` even if we couldn't sync the registry due to being unable to fetch the public key. Instead we return a proper error.
* added telemetry that tracks the initial sync failures. 
* removed some `FIXME` comments and created an [improvement ticket](https://linear.app/dfinity/issue/OBS-476/msd-improvements) where we can properly track the work needed


pietrodimarco-dfinity (approved):
Thank you! :) Most of the suggestions fall under the already existing story https://dfinity.atlassian.net/browse/OB-506 maybe we should link to this.

PR #192, opened 2024-02-13
Persist definitions to disk on add/remove

pietrodimarco-dfinity:
* Change the way we persist definitions to disks to rewrite them every time we add/remove definitions from the server


sasa-tomic (approved):
LGTM! That was super quick!

PR #193, opened 2024-02-14
(Remove backend server from CLI) - 01 - Do not use backend server for hostos-rollout cmd CLI

pietrodimarco-dfinity:
This is the first of a series of PRs that aim to remove backend server usage from the CLI code.

* Migrate `hostos_rollout` from backend to CLI
* Refactor code around this change


sasa-tomic (approved):
LGTM

PR #198, opened 2024-02-18
Add attributes TargetGroup filter for MSD

pietrodimarco-dfinity:
* Added attributes TargetGroup filter to support Prometheus target filter


DFINITYManu (changes requested):
Please document this in the readme.

PR #206, opened 2024-02-19
feat(dre): extending voting script to warn of non-latest release

NikolaMilosa:
Extended the functionality of voting to fetch the latest releases in each iteration and to warn the user if he is updating to a version which is not the newest. User can choose to opt out of voting and then the proposal will not be voted and the user will not be asked again during the same script run. This aims to increase control and add insight into the voting process.

### DIFF
--- a/rs/cli/src/general.rs
+++ b/rs/cli/src/general.rs
@@ -1,4 +1,9 @@
+use candid::Decode;
+use dialoguer::Confirm;
 use ic_base_types::{CanisterId, PrincipalId};
+use ic_management_backend::registry::RegistryState;
+use ic_management_types::Network;
+use registry_canister::mutations::do_update_subnet_replica::UpdateSubnetReplicaVersionPayload;
 use spinners::{Spinner, Spinners};
 use std::{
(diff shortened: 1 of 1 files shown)


rs/cli/src/general.rs
```diff
@@ -51,12 +73,41 @@ pub(crate) async fn vote_on_proposals(
...
+                },
+                _ => {
+                    warn!("Proposal's content wasn't correct. Skipping...");
+                    continue;
+                }
+            };
+
+            if parsed.replica_version_id != *latest_version
+                && !Confirm::new()
+                    .with_prompt(format!("CAUTION! You are about to vote on a proposal that will update the subnet '{}' to a non-latest version! Do you want to continue?", parsed.subnet_id.0.to_string().split('-').collect::<Vec<_>>().first().unwrap()))
```
sasa-tomic, line 96: This could be a bit annoying if we're rolling out two versions at the same time.
But I suppose "remembering" such version for the duration of the single run could be a good compromise.
WDYT?

PR #213, opened 2024-02-27
fix(log-fetcher): binary fields that end with `\n` are not parsed correctly

NikolaMilosa:


rs/log-fetcher/src/journald_parser.rs
```diff
@@ -12,7 +12,81 @@ pub struct JournalEntry {
...
+
+pub fn parse_journal_entries_new(body: &[u8]) -> Vec<JournalEntry> {
+    let mut entries = Vec::new();
+    let mut current_entry = Vec::new();
+    let mut current_line = Vec::new();
+
+    let mut first_found = LineStatus::NotStarted;
+
+    let mut iter = body.iter();
+    while let Some(byte) = iter.next() {
```
sasa-tomic, line 31: Some comment on how the state machine / input parsing works on a high level would be awesome.

sasa-tomic (approved):
LGTM, thanks!

pietrodimarco-dfinity (approved):

PR #222, opened 2024-02-29
Add Boundary Nodes persist

pietrodimarco-dfinity:
* Add persistence for boundary nodes
* Test persist function


NikolaMilosa (approved):
Thanks!

rs/ic-observability/multiservice-discovery/src/server_handlers/add_boundary_node_to_definition_handler.rs
```diff
@@ -49,7 +50,12 @@ pub(super) async fn add_boundary_node(
     };
 
     match running_definition.add_boundary_node(bn).await {
-        Ok(()) => ok(binding.log, format!("Definition {} added successfully", name)),
+        Ok(()) => {
+            if let Err(e) = binding.supervisor.persist_defs(&mut definitions).await {
+                warn!(binding.log, "Error while peristing definitions to disk '{}'", e);
```
sasa-tomic, line 55: Hmm this raises an interesting question. Should we persist a flag somewhere in this case that the definition could not be persisted, to retry persisting later on?

sasa-tomic (approved):
LGTM

rs/ic-observability/multiservice-discovery/src/server_handlers/add_boundary_node_to_definition_handler.rs
pietrodimarco-dfinity: Thx @sasa-tomic ! Yes I believe what we could go that way maybe retrying last time the write to disk when the shutdown happens. wdyt?

PR #224, opened 2024-03-01
feat(rc): started work on `rollout-controller`

NikolaMilosa:
This PR starts the work for the `rollout-controller` which will be used to automate release process. 
Done:

- [x] Sync with registry
- [x] Sync with github
- [ ] ~Notify listeners that `release-index.yaml` changed~ Refactored

Todo:

- [ ] A lot.


rs/ic-management-types/src/lib.rs
```diff
@@ -563,6 +564,16 @@ pub enum Network {
     Url(url::Url),
 }
 
+impl Debug for Network {
```
LittleChimera, line 567: Network already implements Display so you should be able to use it instead of rewriting.

NikolaMilosa: Since `Cli` derives off of `Debug` I need it explicitly

LittleChimera: it's ok, just call .display() instead of providing manual values.

LittleChimera (approved):

PR #230, opened 2024-03-05
Fix firewall rule management.

DFINITYManu:
* Fix typo in proposal to remove firewall rules.
* Fix signature of remove command being called.
* Batch multiple operations so we can at least remove / modify / add multiple rules in one proposal.
* Add standard title, summary and motivation parameters to firewall management.

Before we do the batching, we sort the rules by modification slot
then we reverse the order of the rules list, and then we do the 
batching.  This ensures that the modifications happen in an order
that will not conflict between each other (what does it mean "add 
rule in position 3" if rule in position 2 was removed prior to 
that?).

### DIFF
--- a/rs/cli/src/ic_admin.rs
+++ b/rs/cli/src/ic_admin.rs
@@ -18,14 +18,16 @@ use log::{error, info, warn};
 use prost::Message;
 use regex::Regex;
 use reqwest::StatusCode;
+use serde::Serialize;
 use sha2::{Digest, Sha256};
 use std::collections::BTreeMap;
+use std::fmt;
 use std::fs::File;
 use std::io::{Read, Write};
 use std::os::unix::fs::PermissionsExt;
 use std::process::Stdio;
 use std::time::Duration;
-use std::{path::Path, process::Command};
+use std::{fmt::Display, path::Path, process::Command};
 use strum::Display;
 use tempfile::NamedTempFile;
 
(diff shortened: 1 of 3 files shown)


NikolaMilosa (approved):
Ah sorry. Great catch!

DFINITYManu:
@sasa-tomic @NikolaMilosa PTAL.

rs/cli/src/ic_admin.rs
```diff
@@ -679,112 +785,116 @@ must be identical, and must match the SHA256 from the payload of the NNS proposa
...
-        if added_entries.len() + updated_entries.len() + removed_entries.len() > 1 {
+        let diff = serde_json::to_string_pretty(&reverse_sorted).unwrap();
+        info!("Pretty printing diff:\n{}", diff);
+        /*if reverse_sorted.len() > 1 {
             return Err(anyhow::anyhow!(
-                "Cannot apply more than 1 change at a time due to hash changes"
+                "Cannot currently apply more than 1 change at a time due to hash changes"
             ));
-        }
+        }*/
```
sasa-tomic, line 810: remove?

sasa-tomic (approved):
LGTM, but the complexity grew substantially compared to the original PR. I don't see an obvious issue, but this is a critical piece of configuration so IMHO it would be good to at least run this on a testnet to verify that the code would work as expected.

rs/cli/src/ic_admin.rs
DFINITYManu: I'll add that this change does not impact any critical piece of configuration.  This change merely modifies how we call `ic-admin` and what data we pass onto it.  I want to be very clear about the scope of this change:

There are two ways in which someone could maliciously or accidentally operate this tool in which it could hypothetically break the IC.  The operator -- or the tool -- could enter an invalid payload (junk), or he could enter a vexatious yet valid payload (e.g. "block all connections from all IPs")

W.r.t invalid payloads:

* The correct place to establish safeguards as to what critical data and code is valid is on the server side (the replica), and the safeguards are already established there plus they are well-tested.
* The second place where safeguards should be in place is `ic-admin` itself (since `dre` merely invokes it here).  I am happy to report that I have personally tested `ic-admin` refuses to load invalid payloads, so the safeguards are already in place.

W.r.t vexatious payloads:

Nothing we can do from the perspective of maintainers of `dre`.  If someone sends a proposal that wrecks the IC, and the proposal gets voted in, then that's not a responsibility of the authors of the `dre` tool.

We could hypothetically add some heuristic that prevents extreme breakage (detect overly zealous block rules, like `rm -rf /` does today).  That, however, would not make this specific change any safer -- it is an engineering decision entirely orthogonal to this change.

DFINITYManu:
After merge this will be tested in staging.

PR #287, opened 2024-04-03
Small fixes for release controller

LittleChimera:


release-controller/reconciler.py
```diff
@@ -278,22 +278,16 @@ def main():
 
 def oneoff():
```
sasa-tomic, line 279: Hmm, what does this function do, and what do changes do?
Can you add a comment (docstring) to the function, with some explanation of the function purpose?

sasa-tomic (approved):
The changes look harmless but to be honest I don't yet understand this part of the code. Approving to unblock.

release-controller/reconciler.py
LittleChimera:
```suggestion
# use this as a template in case you need to manually submit a proposal
def oneoff():
```

sasa-tomic (approved):

PR #309, opened 2024-04-09
Implement filtering for the discovery service.

DFINITYManu:
Filtering by datacenter ID, node provider, subnet, and IC name are implemented.

### DIFF
--- a/rs/cli/src/clients.rs
+++ b/rs/cli/src/clients.rs
@@ -160,19 +160,19 @@ mod tests {
             .expect("failed to create staging network");
         assert_eq!(
             DashboardBackendClient::new(&mainnet, false).url.to_string(),
-            "https://dashboard.internal.dfinity.network/api/proxy/registry/mainnet/"
+            "https://dashboard.internal.dfinity.network/api/proxy/registry/mainnet"
         );
         assert_eq!(
             DashboardBackendClient::new(&staging, false).url.to_string(),
-            "https://dashboard.internal.dfinity.network/api/proxy/registry/staging/"
+            "https://dashboard.internal.dfinity.network/api/proxy/registry/staging"
         );
         assert_eq!(
             DashboardBackendClient::new(&mainnet, true).url.to_string(),
-            "http://localhost:17000/api/proxy/registry/mainnet/"
+            "http://localhost:17000/api/proxy/registry/mainnet"
         );
         assert_eq!(
             DashboardBackendClient::new(&staging, true).url.to_string(),
-            "http://localhost:17000/api/proxy/registry/staging/"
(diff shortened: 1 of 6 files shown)


NikolaMilosa (approved):
Lgtm! Thanks

rs/ic-observability/multiservice-discovery/src/definition.rs
```diff
@@ -669,3 +670,72 @@ impl DefinitionsSupervisor {
...
+        };
+        o && n && d && s
+    }
+
+    pub fn matches_boundary_node(&self, b: &BoundaryNode) -> bool {
+        // self.ic_name is explicitly excluded here.
+        // Call self.matches_ic().
+        if self.operator_id.is_some() || self.node_provider_id.is_some() || self.subnet_id.is_some() {
+            return true;
+        };
```
sasa-tomic, line 714: Is this what we want? If any of these fields are provided, the BN filtering will "match" without and further checks.
What if NP + DC filter is provided? The function will return all BNs.

DFINITYManu: Whoa, what a wonderful catch.  It turns out it's meant to be return false.

rs/cli/src/clients.rs
```diff
@@ -160,19 +160,19 @@ mod tests {
             .expect("failed to create staging network");
         assert_eq!(
             DashboardBackendClient::new(&mainnet, false).url.to_string(),
-            "https://dashboard.internal.dfinity.network/api/proxy/registry/mainnet/"
```
DFINITYManu, line 163: These tests were broken.

rs/ic-observability/multiservice-discovery/src/server_handlers/export_prometheus_config_handler.rs
```diff
@@ -80,11 +47,14 @@ pub fn serialize_definitions_to_prometheus_config(definitions: BTreeMap<String,
...
-pub(super) async fn export_prometheus_config(State(binding): State<Server>) -> Result<String, (StatusCode, String)> {
+pub(super) async fn export_prometheus_config(
+    State(binding): State<Server>,
+    filters: Query<TargetFilterSpec>,
+) -> Result<String, (StatusCode, String)> {
     let definitions = binding.supervisor.definitions.lock().await;
-    let (targets_len, text) = serialize_definitions_to_prometheus_config(definitions.clone());
+    let (targets_len, text) = serialize_definitions_to_prometheus_config(definitions.clone(), filters.0);
     if targets_len > 0 {
-        ok(binding.log, text)
```
DFINITYManu, line 87: This was logging text unnecessarily.

PR #316, opened 2024-04-12
Return only host/port.

DFINITYManu:
We were returning http:// or https:// along with the address, as well as a /path at the end. This is not supported by Prometheus and therefore this PR removes it.  With this, Prometheus correctly loads the targets, at the cost of the client needing to hard code the scheme and the path in the SD config.

**Note: this breaks compatibility.  If any part of our stack is currently relying on this broken format, we will have to figure out a way to safely roll this out.**

Prometheus cannot natively cope with protocol and path in the address.  This can be worked around using relabeling rules:

```
  - # Strip the scheme from the multiservice-discovery-generated address.
    source_labels:
    - __address__
    target_label: __address__
    regex: (.*)://(.*)
    replacement: ${2}
    action: replace
  - # Strip the metrics path from the multiservice-discovery-generated address.
    source_labels:
    - __address__
    target_label: __address__
    regex: (.*?)/(.*)
    replacement: ${1}
    action: replace
```

(I tested the above locally.)

The problem with this PR removing the protocol and the path in the returned address is that, without this information in the target, we cannot configure a single discovery service in our IC Metrics stack.  We would have to configure multiple ones (one per target), and drop the targets we don't want using rewrite rules.  Alternatively, we could use regexes to *fish out* the protocol and the path, and rewrite the __scheme__ and __path__ magic variables during relabeling.  This seems exceedingly hacky, but I'm not seeing anything outside these two solutions.

This is what I am currently doing in IC-O-Stack, as we have a fixed list of target URLs we scrape.

If we are OK with going full rewrite using regexes and relabeling rules, then this PR is not needed.

--------

https://github.com/dfinity/dre/pull/315 is required below this PR.

### DIFF
--- a/rs/ic-observability/multiservice-discovery-shared/src/builders/prometheus_config_structure.rs
+++ b/rs/ic-observability/multiservice-discovery-shared/src/builders/prometheus_config_structure.rs
@@ -46,7 +46,7 @@ pub fn map_target_group(target_groups: Vec<TargetDto>) -> Vec<PrometheusStaticCo
             let mut ret = vec![];
             for job in &tg.jobs {
                 ret.push(PrometheusStaticConfig {
-                    targets: tg.targets.iter().map(|sa| job.url(*sa, false)).collect(),
+                    targets: tg.targets.iter().map(|sa| job.host_port(*sa, false)).collect(),
                     labels: {
                         BTreeMap::from([
                             (IC_NAME.into(), tg.ic_name.clone()),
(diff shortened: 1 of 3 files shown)


sasa-tomic (approved):
I prefer clean design if it's possible. I don't think that we rely on the prometheus target list for anything yet.

NikolaMilosa:
> I prefer clean design if it's possible. I don't think that we rely on the prometheus target list for anything yet.

AFAIK, staging and testnets are being scraped via this url so this has to be tested with victoria.

DFINITYManu:
@NikolaMilosa hoping the trip back was great!  Too short a visit, maybe next time it can be longer.

Can I have a search string or path name to look at the Victoria K8s config?

DFINITYManu:
Turns out we can't do this and stay compatible with Victoria.

image

It expects an URL with host and port, instead of entries for the scheme and host:port.

This sucks.  We'll have to stick in the IC observability stack to the trick that uses relabeling instead.

DFINITYManu:
This won't be necessary in the end.

PR #357, opened 2024-05-06
feat(dre): adding api boundary nodes to registry dump command

NikolaMilosa:
Alongside this there are some minor changes related to aligning with a newer version of IC repo. Those were needed since there were renamings in the upstream proposal paylod struct names


LittleChimera (approved):

sasa-tomic (approved):
Nice, thanks!

PR #362, opened 2024-05-08
feat: adding using of public dashboard for mainnet for health statuses

NikolaMilosa:
Improvement towards using `dre` tool without vpn


sasa-tomic (approved):
Awesome!!!

PR #372, opened 2024-05-13
feat(dre): listing proposals

NikolaMilosa:
Currently supports:
1. Listing pending proposals => maps to [`get_pending_proposals`](https://dashboard.internetcomputer.org/canister/rrkah-fqaaa-aaaaa-aaaaq-cai#get_pending_proposals)
2. Listing proposal with filtering => maps to [`list_proposals`](https://dashboard.internetcomputer.org/canister/rrkah-fqaaa-aaaaa-aaaaq-cai#get_pending_proposals)

### DIFF
--- a/rs/cli/src/cli.rs
+++ b/rs/cli/src/cli.rs
@@ -154,6 +154,9 @@ pub enum Commands {
         #[clap(long, default_value = None, required = true)]
         summary: Option<String>,
     },
+
+    /// Proposal Listing
+    Proposals(proposals::Cmd),
 }
 
 impl Default for Commands {
@@ -421,5 +424,21 @@ pub mod nodes {
             #[clap(long, aliases = ["summary"])]
             motivation: Option<String>,
         },
+    }
+}
+
+pub mod proposals {
+    use super::*;
+
+    #[derive(Parser, Clone)]
+    pub struct Cmd {
+        #[clap(subcommand)]
+        pub subcommand: Commands,
+    }
+
+    #[derive(Subcommand, Clone)]
+    pub enum Commands {
+        /// Get list of pending proposals
+        Pending,
     }
 }
--- a/rs/cli/src/cli.rs
+++ b/rs/cli/src/cli.rs
@@ -440,5 +440,49 @@ pub mod proposals {
     pub enum Commands {
         /// Get list of pending proposals
         Pending,
+
+        /// Get list of filtered proposals
+        List {
+            /// Limit on the number of \[ProposalInfo\] to return. If no value is
+            /// specified, or if a value greater than 100 is specified, 100
+            /// will be used.
+            #[clap(long, default_value = "100")]
+            limit: u32,
+            /// If specified, only return proposals that are strictly earlier than
+            /// the specified proposal according to the proposal ID. If not
+            /// specified, start with the most recent proposal.
+            #[clap(long)]
+            before_proposal: Option<u64>,
+            /// Exclude proposals with a topic in this list. This is particularly
+            /// useful to exclude proposals on the topics TOPIC_EXCHANGE_RATE and
+            /// TOPIC_KYC which most users are not likely to be interested in
+            /// seeing.
+            #[clap(long)]
+            exclude_topic: Vec<i32>,
+            /// Include proposals that have a reward status in this list (see
+            /// \[ProposalRewardStatus\] for more information). If this list is
+            /// empty, no restriction is applied. For example, many users listing
+            /// proposals will only be interested in proposals for which they can
+            /// receive voting rewards, i.e., with reward status
+            /// PROPOSAL_REWARD_STATUS_ACCEPT_VOTES.
+            #[clap(long)]
+            include_reward_status: Vec<i32>,
+            /// Include proposals that have a status in this list (see
+            /// \[ProposalStatus\] for more information). If this list is empty, no
+            /// restriction is applied.
+            #[clap(long)]
+            include_status: Vec<i32>,
+            /// Include all ManageNeuron proposals regardless of the visibility of the
+            /// proposal to the caller principal. Note that exclude_topic is still
+            /// respected even when this option is set to true.
+            #[clap(long)]
+            include_all_manage_neuron_proposals: Option<bool>,
+            /// Omits "large fields" from the response. Currently only omits the
+            /// `logo` and `token_logo` field of CreateServiceNervousSystem proposals. This
+            /// is useful to improve download times and to ensure that the response to the
+            /// request doesn't exceed the message size limit.
+            #[clap(long)]
+            omit_large_fields: Option<bool>,
+        },
     }
 }
--- a/rs/cli/src/main.rs
+++ b/rs/cli/src/main.rs
@@ -343,10 +343,7 @@ async fn async_main() -> Result<(), anyhow::Error> {
                     let nns_url = target_network.get_nns_urls().first().expect("Should have at least one NNS URL");
                     let client = GovernanceCanisterWrapper::from(CanisterClient::from_anonymous(nns_url)?);
                     let proposals = client.list_proposals(ListProposalInfo {
-                        before_proposal: match before_proposal {
-                            None => None,
-                            Some(p) => Some(ProposalId { id: *p })
-                        },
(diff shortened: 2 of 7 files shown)


rs/cli/src/main.rs
```diff
@@ -327,6 +330,32 @@ async fn async_main() -> Result<(), anyhow::Error> {
                     ..Default::default()
                 }, cli_opts.simulate).await
             }
+            cli::Commands::Proposals(p) => match &p.subcommand {
+                cli::proposals::Commands::Pending => {
+                    let nns_url = target_network.get_nns_urls().first().expect("Should have at least one NNS URL");
+                    let client = GovernanceCanisterWrapper::from(CanisterClient::from_anonymous(nns_url)?);
+                    let proposals = client.get_pending_proposals().await?;
+                    let proposals = serde_json::to_string(&proposals).map_err(|e| anyhow::anyhow!("Couldn't serialize to string: {:?}", e))?;
```
sasa-tomic, line 338:
```suggestion
                    let proposals = serde_json::to_string_pretty(&proposals).map_err(|e| anyhow::anyhow!("Couldn't serialize to string: {:?}", e))?;
```

rs/cli/src/main.rs
```diff
@@ -327,6 +330,32 @@ async fn async_main() -> Result<(), anyhow::Error> {
...
+                    let proposals = client.list_proposals(ListProposalInfo {
+                        before_proposal: before_proposal.as_ref().map(|p| ProposalId { id: *p }),
+                        exclude_topic: exclude_topic.clone(),
+                        include_all_manage_neuron_proposals: *include_all_manage_neuron_proposals,
+                        include_reward_status: include_reward_status.clone(),
+                        include_status: include_status.clone(),
+                        limit: *limit,
+                        omit_large_fields: *omit_large_fields
+                    }).await?;
+                    let proposals = serde_json::to_string(&proposals).map_err(|e| anyhow::anyhow!("Couldn't serialize to string: {:?}", e))?;
```
sasa-tomic, line 354:
```suggestion
                    let proposals = serde_json::to_string_pretty(&proposals).map_err(|e| anyhow::anyhow!("Couldn't serialize to string: {:?}", e))?;
```

sasa-tomic (approved):
Whoa, fantastic!!! :rocket:

rs/cli/src/main.rs
NikolaMilosa: Wouldn't it be better to pipe it to `jq` if you need pretty string? The intention for this is just to dump the proposals (similar as to `dump-registry`) and pipe consume it with something else. I doubt it that this will be used from an end user being an engineer. WDYT?

rs/cli/src/cli.rs
```diff
@@ -423,3 +426,63 @@ pub mod nodes {
...
+            /// the specified proposal according to the proposal ID. If not
+            /// specified, start with the most recent proposal.
+            #[clap(long)]
+            before_proposal: Option<u64>,
+            /// Exclude proposals with a topic in this list. This is particularly
+            /// useful to exclude proposals on the topics TOPIC_EXCHANGE_RATE and
+            /// TOPIC_KYC which most users are not likely to be interested in
+            /// seeing.
+            #[clap(long)]
+            exclude_topic: Vec<i32>,
```
DFINITYManu, line 461: This needs to be a list of strings, not a single string.  It should also be named in the plural to indicate it's a list.  Not sure how clap handles lists.

DFINITYManu (changes requested):

rs/cli/src/cli.rs
NikolaMilosa: Not the best solution right now. If you take a look at how `ListProposalInfo` struct is formatted it also expects a `Vec<i32>`. On top of that having strings means that we would need to keep those in sync and the basic rename would break while in reality it shouldn't. I would like to keep this api as is, as much as possible because I think this would be used wrapped from another source (like Airflow) where you would be able to map strings to concrete integers. I would create a separate command which would be used from an engineer. It could include the mappings you suggested, and other fancy things. Your request would break that.

NikolaMilosa: Source: https://github.com/dfinity/ic/blob/master/rs/nns/governance/src/gen/ic_nns_governance.pb.v1.rs#L2715

rs/cli/src/main.rs
DFINITYManu: A nice output is preferable to a noisy blob.  We _can_ pipe things to other things, but I would prefer not to _have_ to do that.

rs/cli/src/cli.rs
DFINITYManu: > If you take a look at how ListProposalInfo struct is formatted it also expects a Vec<i32>. 

Correct, which means the tool has to take the list of strings specified, convert them to the right integers, and assemble the right `Vec` on behalf of the user.

The software exists to serve the needs of the people, not the other way around.

rs/cli/src/main.rs
NikolaMilosa: It will be a noisy blob anyway, JSON serializers serialize array's as each element in a new line. Since the payload is an array of bytes you get a really large and ugly looking array (which means nothing) and it takes a lot of the terminal. I think its better to have a default and consume it if you really need it to. Which I doubt anyone will without further filtering and selecting relevant data (also with jq).

rs/cli/src/cli.rs
NikolaMilosa: For this command specifically. I think its best if we left it mapped exactly as is to the upstream with little to no mappings. "Software" part should come in a different command which will be more userfriendly and tailored to the need. This command should be there and integrated with different tools. If i add mapping here from strings to int you will have to add a mapping from int to string in Airflow (and you don't have that right now, you have enums) meaning we achieved nothing and have to maintain two maps in sync. This is just a map from the call of a canister allowed to the user via cli. Your request will be granted in a different command.

DFINITYManu: OK.

rs/cli/src/main.rs
sasa-tomic: The question is really who is the expected user of this output?
A machine doesn't care if it's pretty or not, and we don't care (that much) about the bandwidth either. If the bandwidth is an issue, let's use json+gzip. But I don't think it is something we need to do.
A human would care if the output is pretty or not, and would have to use jq every time, if they would ever use the output. Adding this requirement of the jq doesn't make much sense IMHO.
Personally, I do expect the output to be viewed by user(s), at least sometimes.

rs/cli/src/cli.rs
sasa-tomic: Can we use `strum` to map enums to strings and back?

rs/cli/src/main.rs
NikolaMilosa: I would prefer to create a different command for the users since this is just a map from gov canister. This was created with the vision of replacing Airflow's dependency to public dashboard. I really doubt that user needs all of this data that is here.

rs/cli/src/cli.rs
NikolaMilosa: Probably yes, but it doesn't make sense because neither will the input from user be that enum, nur will the governance canister accept an enum

sasa-tomic: Aha, there is already a [conversion to string](https://sourcegraph.com/github.com/dfinity/ic@09ce93457a49ba4ca452cf9e4d2d46dd54b3d146/-/blob/rs/nns/governance/src/gen/ic_nns_governance.pb.v1.rs?L3372-3390)
and [conversion from string](https://sourcegraph.com/github.com/dfinity/ic@09ce93457a49ba4ca452cf9e4d2d46dd54b3d146/-/blob/rs/nns/governance/src/gen/ic_nns_governance.pb.v1.rs?L3392-3411)

rs/cli/src/main.rs
sasa-tomic: I'm okay if we add a different user-facing command. Do you already have plans on how that will look like?

NikolaMilosa: Yes, as of now it will include:
1. greater limit (you could request more than 100 proposals)
2. would work on `including` rather than `excluding`
3. would have way less data mapped to more friendly interface since now we just return what we map from gov canister. (Things like binary payload and so on)

PR #388, opened 2024-05-17
Cleanup release index

LittleChimera:


NikolaMilosa (approved):
Noice

DFINITYManu (changes requested):
why?  no need to delete old records, no?  if the software has a bug, the software should be fixed.

sasa-tomic:
Actually seems like the unassigned nodes did not get updated as they should have been and therefore we had old revisions being "active" and constantly rechecked.
Instead of merging this PR we should probably fix the regular updates of the unassigned nodes. There is now a separate proposal that we can vote in, that updates unassigned nodes. It will be used by default in the DRE tool once https://github.com/dfinity/dre/pull/411 is merged.

PR #398, opened 2024-05-21
fix(release-controller): [DRE-166] use requests instead of urllib, to handle timeouts

sasa-tomic:
source: https://stackoverflow.com/questions/32763720/timeout-a-file-download-with-python-urllib


NikolaMilosa (approved):
✅

PR #399, opened 2024-05-21
feat(observability): Automatically update k8s vector container versions

sasa-tomic:
Update the vector mainnet and testnet container versions in the k8s repo, when the DRE repo version changes.


sasa-tomic:
> It looks good, but if we are going this route, maybe we should also add the `bases/apps/service-discovery`?

Sounds good, let's do that!

PR #400, opened 2024-05-21
chore(ci): changing main pipeline to run on self hosted runners

NikolaMilosa:


.github/workflows/main.yaml
```diff
@@ -21,7 +21,9 @@ concurrency:
 
 jobs:
   bazel:
-    runs-on: ubuntu-20.04
```
sasa-tomic, line 24: What's the OS version of the custom runners?

.github/workflows/build/action.yaml
```diff
@@ -55,25 +55,28 @@ runs:
         branch: ${{ github.head_ref }}
         ssh: true
 
-    - name: Dry run tarballs
```
sasa-tomic, line 58: We could still run this on GH runners, but conditioning with [if ](https://docs.github.com/en/actions/learn-github-actions/contexts#runner-context)runner.name[ != something](https://docs.github.com/en/actions/learn-github-actions/contexts#runner-context).

Or do you have a plan on how to run this on the custom runners in the (near) future?

.github/workflows/main.yaml
```diff
@@ -125,11 +127,3 @@ jobs:
           git config --global user.email "actions@github.com"
           git config --global user.name "GitHub Actions"
           bazel run "//:mkdocs" -- gh-deploy --force
-
-      ########################################
-      # Clean up runner
-      ########################################
-      - uses: ./.github/workflows/manage-runner-post
```
sasa-tomic, line 132: We could skip this step if running on our custom runners, instead of deleting the workflow completely. It was helping a lot on upstream GH runners and we might want to be able to run jobs there as well.

.github/workflows/manage-runner-post/action.yaml
```diff
@@ -1,64 +0,0 @@
...
-    required: true
-
-runs:
-  using: composite
-  steps:
-    ########################################
-    # Stop bazel-remote cache
-    ########################################
-    - name: "💾 Stop bazel-remote s3 cache"
-      run: ./bin/bazel-cache-s3-stop.sh
```
sasa-tomic, line 16: If we decide to delete the workflows in the end, we don't need these scripts either.

.github/workflows/manage-runner-post/action.yaml
```diff
@@ -1,64 +0,0 @@
...
-      shell: bash
-
-    ########################################
-    # Optimize bazel cache by hard-linking duplicate files
-    ########################################
-    - name: "🧹 Optimize bazel cache directory before uploading"
-      if: ${{ github.ref == 'refs/heads/main' }}
-      run: ./bin/optimize-bazel-cache.sh
```
sasa-tomic, line 24: This could also be deleted if we delete the workflow

.github/workflows/main.yaml
NikolaMilosa:
```bash
k exec dre-runner-custom-jlm9w-runner-65lbr -c runner -- cat /etc/os-release 
PRETTY_NAME="Ubuntu 22.04.4 LTS"
NAME="Ubuntu"
VERSION_ID="22.04"
VERSION="22.04.4 LTS (Jammy Jellyfish)"
VERSION_CODENAME=jammy
ID=ubuntu
ID_LIKE=debian
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
UBUNTU_CODENAME=jammy
```

.github/workflows/build/action.yaml
NikolaMilosa: It will be super hard to manage cache between self hosted runners and github runners. I will look into fixing this. I know what the issue is just have to see what are all the possibilities of fixing it

.github/workflows/main.yaml
NikolaMilosa: If we step in the way of self hosted runners i doubt it that we will want to revert back. Maintaining legacy pipeline will be hard and that pipeline will just grow and grow with unreadable if/elses. I think that if we want to make a turn towards self-hosted we shouldn't look into maintaining both

.github/workflows/build/action.yaml
sasa-tomic: I don't think that's necessary. We can let GH rebuild the cache on master every now and then (e.g. once per week), that should be enough IMHO.

.github/workflows/main.yaml
sasa-tomic: To avoid glibc and other lib issues (again) we should use the oldest OS version possible. Can we use Ubuntu 20.04?

docker/runner.Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM ghcr.io/actions/actions-runner:2.316.1
```
sasa-tomic, line 1: Switching to Ubuntu 20.04 seems to be as easy as:

```suggestion
FROM ghcr.io/actions/actions-runner:v2.316.1-ubuntu-20.04
```

.github/workflows/main.yaml
NikolaMilosa: Yeah, i will test it

docker/runner.Dockerfile
sasa-tomic, line 1: Here are a few docs for building alternative images based on older OS:

https://baccini-al.medium.com/how-to-containerize-a-github-actions-self-hosted-runner-5994cc08b9fb

https://gist.github.com/deptno/20c999efd4e00ee5e5fb9861db4a8ee1

I haven't tried any of them, so take it with a grain of salt.

NikolaMilosa: It doesn't have to have actions-runner in it. I took it because i thought it had most of the things we need. We can really base everything from ubuntu:20.04

sasa-tomic (approved):
Fantastic, thanks!

docker/runner.Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+ENV TZ=UTC
+ENV DEBIAN_FRONTEND=noninteractive
+ENV RUNNER_UID=1001
+
+RUN apt-get update && \
+    apt-get upgrade -y && \
+    apt-get install ca-certificates curl git-all gcc g++ clang pkg-config make sudo docker.io build-essential -y
+
+RUN mkdir -p openssl && \
+    curl -o openssl/openssl-1.1.1w.tar.gz -L https://www.openssl.org/source/old/1.1.1/openssl-1.1.1w.tar.gz && \
```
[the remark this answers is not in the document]
NikolaMilosa: Openssl is not even installed on ubuntu:20.04

LittleChimera (approved):

PR #422, opened 2024-05-24
Consolidate teams that get to review the GuestOS release notes into a…

DFINITYManu:
… single list.

Previousy this list was maintained redundantly in two different places, which led to the document containing calls for teams to review the document whose team members did not get a Slack message.


sasa-tomic (approved):
Whoa, that's awesome! Great idea!

PR #435, opened 2024-05-29
fix: Add a time for all agent-rs (canister) clients

sasa-tomic:
We observed that the NNS proposal notification bot got stuck while checking for failed proposals.

### DIFF
--- a/rs/ic-canisters/src/lib.rs
+++ b/rs/ic-canisters/src/lib.rs
@@ -15,6 +15,7 @@ use serde::Deserialize;
 use std::path::PathBuf;
 use std::str::FromStr;
 use std::sync::Mutex;
+use std::time::Duration;
 use url::Url;
 
 pub mod governance;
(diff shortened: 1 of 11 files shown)


rs/ic-canisters/src/lib.rs
```diff
@@ -83,13 +84,17 @@ impl IcAgentCanisterClient {
...
+            .use_rustls_tls()
+            .timeout(Duration::from_secs(30))
+            .build()
+            .expect("Could not create HTTP client.");
+        let agent = Agent::builder()
+            .with_identity(identity)
+            .with_transport(ReqwestTransport::create_with_client(url, client)?)
+            .with_verify_query_signatures(false)
+            .build()?;
+        Ok(Self { agent })
```
NikolaMilosa, line 97: Looks like we reuse this functionality a lot. Can we call it in other parts of code? Looks like it is basically the same code

NikolaMilosa (approved):
LGTM

rs/ic-canisters/src/lib.rs
sasa-tomic: Maybe we could create our own wrapper for this, but it would actually be better to have https://github.com/dfinity/agent-rs/pull/277 merged in agent-rs. Then we wouldn't need this boilerplate code. So I'll try to push for that instead.

PR #438, opened 2024-05-30
Allow missing team release controller

pietrodimarco-dfinity:


sasa-tomic (approved):
Thank you!

scripts/release-notes.py
```diff
@@ -60,6 +60,7 @@
     "prodsec": "Prodsec",
     "runtime-owners": "Runtime",
     "trust-team": "Trust",
+    "utopia": "Utopia"
```
DFINITYManu, line 63: We may want to replace some of these with a heuristic that removes -owners and -team, then capitalizes, and keep this for exceptional cases when this isn't true.

PR #442, opened 2024-05-31
feat(dre): Convenience function for dumping registry records for incorrect rewards

sasa-tomic:
This same output can be achieved with some `jq` gymnastics, but this PR facilitates the work.

Also, added the ability to write the output to a file. Default is stdout, as before.


NikolaMilosa (approved):
Thanks!

PR #455, opened 2024-06-05
chore(dre): Rotate runners local cache

pietrodimarco-dfinity:
This implements a round-robin cache allocation strategy for the runners.
It also enables concurrent CI job execution using cache stored on a shared local volume.


.github/workflows/manage-runner-pre/action.yaml
```diff
@@ -4,6 +4,28 @@ description: Reusable action for setting up the github runner
...
   using: composite
   steps:
+    - name: "💾 Rotate runner cache"
+      if: ${{ startsWith(runner.name, 'dre-runner-custom') }}
+      shell: bash
+      run: |
+        set -exuo pipefail
+
+        RUNNERS_AVAILABLE=4
+        CURRENT_CACHE_FILE="/home/runner/.cache_pool/current_cache.txt"
```
LittleChimera, line 14: is there a risk of race condition here?

pietrodimarco-dfinity: Good catch yes true. Let me fix it

PR #456, opened 2024-06-05
feature(dre): add all methods to manage the API boundary nodes

r-birkner:


rs/cli/src/main.rs
```diff
@@ -354,6 +355,60 @@ async fn async_main() -> Result<(), anyhow::Error> {
...
+                    runner_instance
+                        .ic_admin
+                        .propose_run(
+                            ic_admin::ProposeCommand::AddApiBoundaryNodes {
+                                nodes: nodes.to_vec(),
+                                version: version.to_string(),
+                            },
+                            ic_admin::ProposeOptions {
+                                title: Some(format!("Add {} API boundary node(s)", nodes.clone().len())),
+                                summary: Some(format!("Add {} API boundary node(s)", nodes.clone().len())),
```
sasa-tomic, line 387: Do we want to have a more explanatory message here, or a link to the forum post if you have one?
I believe not many community members will be familiar how these nodes will be created / updated, etc.

rs/cli/src/main.rs
```diff
@@ -354,6 +355,60 @@ async fn async_main() -> Result<(), anyhow::Error> {
...
+                    runner_instance
+                        .ic_admin
+                        .propose_run(
+                            ic_admin::ProposeCommand::DeployGuestosToSomeApiBoundaryNodes {
+                                nodes: nodes.to_vec(),
+                                version: version.to_string(),
+                            },
+                            ic_admin::ProposeOptions {
+                                title: Some(format!("Update {} API boundary node(s) to {version}", nodes.clone().len())),
+                                summary: Some(format!("Update {} API boundary node(s) to {version}", nodes.clone().len())),
```
sasa-tomic, line 369: Similar comment as below.

rs/cli/src/ic_admin.rs
```diff
@@ -626,7 +626,7 @@ must be identical, and must match the SHA256 from the payload of the NNS proposa
         }
     }
 
-    pub async fn update_unassigned_nodes(&self, nns_subned_id: &String, network: &Network, simulate: bool) -> Result<(), Error> {
+    pub async fn update_unassigned_nodes(&self, nns_subnet_id: &String, network: &Network, simulate: bool) -> Result<(), Error> {
```
sasa-tomic, line 629: Thanks for fixing the typo. Took me a while to realize what changed :D

sasa-tomic (approved):

rs/cli/src/main.rs
r-birkner: I didn't quite understand the difference between summary and motivation. How do you guys use it?

sasa-tomic: We should probably merge the two. The idea was that "summary" should include arbitrary description of the change plus a line with `Motivation: <motivation>` and the "motivation" is just the `<motivation>` part. But we're apparently not very consistent in writing good motivations either way. So we can probably merge the two and simplify the code.

r-birkner: Ok, I added the motivation as required argument for the two commands. I think it actually does make sense to have summary and motivation. I looked at some previous proposals and it looks good.

PR #459, opened 2024-06-06
fix(service-discovery): Delete persisted state file if loading previous contents fails

sasa-tomic:


pietrodimarco-dfinity (approved):
Thx :)

rs/ic-observability/multiservice-discovery/src/main.rs
```diff
@@ -77,8 +77,18 @@ fn main() {
...
-            .unwrap();
+        match rt.block_on(supervisor.load_or_create_defs(metrics.running_definition_metrics.clone())) {
+            Ok(_) => {}
+            Err(e) => {
+                error!(log, "Failed to load or create definitions, deleting the state file.");
+                if let Some(file) = &cli_args.networks_state_file {
+                    if let Err(e) = std::fs::remove_file(file) {
+                        error!(log, "Failed to delete the state file"; "error" => format!("{:?}", e));
+                    }
+                }
```
NikolaMilosa, line 88: Not sure how Pietro implemented it, but by the name it sounds like it is expected that it has a file after that function, so maybe a simple truncating of a file or serializing something valid (empty) in it is the way to go.

sasa-tomic: It checks whether a file exists, otherwise returns `Ok(())`. So let's give this a shot.

PR #466, opened 2024-06-07
chore: changing default behavior if cursor is not present

NikolaMilosa:
Previously, if the `__CURSOR` wasn't present we would start fetching only the newest logs. While convenient it we had some issues where we would lose logs or have to refetch from the start and this would help us achieve that more easily. When the new target is added we will start fetching logs from the start of their journal


sasa-tomic (approved):
LGTM

PR #475, opened 2024-06-10
Make sure that when --nodes is passed to ic-admin, each node is prefi…

DFINITYManu:
…xed with --nodes.

Thanks to @nikola.milosavjevic for the protip.


NikolaMilosa (approved):

r-birkner:
Thanks for spotting the problem and fixing my bugs 🙏

PR #476, opened 2024-06-10
Properly quote ic-admin commands emitted for review.

DFINITYManu:


rs/cli/Cargo.toml
```diff
@@ -69,6 +69,7 @@ tempfile = "3.10.0"
 tokio = { workspace = true }
 url = { workspace = true }
 humantime = { workspace = true }
+shlex = "1.3.0"
```
sasa-tomic, line 72: Can we move this to the workspace, so that we don't need to chase dependencies in all Cargo.toml files in the repo?

sasa-tomic (approved):

rs/cli/Cargo.toml
DFINITYManu: I don't know how to do that.  `cargo add shlex` doesn't work.

DFINITYManu:
```
error: `cargo add` could not determine which package to modify. Use the `--package` option to specify a package. 
available packages: dre, decentralization, ic-management-types, ic-canisters, ic-management-backend, config-writer-common, service-discovery, multiservice-discovery, multiservice-discovery-shared, multiservice-discovery-downloader, node-status-updater, obs-canister-clients, prometheus-config-updater, sns-downloader, log-fetcher, canister-log-fetcher, np-notifications, rollout-controller, slack-notifications
 manuel   properquote ✚ 2  ~  src  dre  101  cargo add shlex --workspace
error: unexpected argument '--workspace' found

  tip: to pass '--workspace' as a value, use '-- --workspace'

Usage: cargo add [OPTIONS] <DEP>[@<VERSION>] ...
       cargo add [OPTIONS] --path <PATH> ...
       cargo add [OPTIONS] --git <URL> ...

For more information, try '--help'.
```

DFINITYManu: Think I found a way.  Not via cli.

DFINITYManu: Done.

sasa-tomic (approved):

PR #479, opened 2024-06-13
fix(dre): Fix the test for the hostos rollout, add api-boundary-nodes

sasa-tomic:
The test somehow didn't have these fields set. We should look into why the CI didn't catch this.


NikolaMilosa (approved):
I see a bug, will fix in another PR

PR #480, opened 2024-06-13
fix(dre): Do not require an HSM for dry runs

sasa-tomic:


rs/cli/src/detect_neuron.rs
```diff
@@ -94,15 +94,29 @@ impl Neuron {
...
+            }
+        };
+        let neuron_id = match auto_detect_neuron_id(self.network.get_nns_urls(), auth).await {
+            Ok(neuron_id) => neuron_id,
+            Err(e) => {
+                if require_auth {
+                    return Err(anyhow::anyhow!(e));
+                } else {
+                    return Ok(vec![]);
+                }
```
NikolaMilosa, line 117: Wouldn't it make things faster if we didn't even try to do anything if we don't require auth? Just return `Ok(vec![])`

sasa-tomic: Yes, that is one option. But then it wouldn't be easy to find out if the HSM parameters are correctly discovered. In that case, the only way to figure out if the HSM works is to try to submit the proposal.

PR #501, opened 2024-06-18
feat(CI): adding job to ensure opentelemetry version

NikolaMilosa:
Found an interesting bug that we really cannot catch... Its caused by `Update dependencies` job (but I am 100% sure that dependabot would cause the same, and maybe has).
Here is a problematic PR with the line: [https://github.com/dfinity/dre/commit/1eda593b16fb3e72e6f102df708b5b69eb502bda#diff-13ee4b2252c9e516a0547f2891aa[…]a71c6d7a1e682c69be97998dfc87eL747](https://github.com/dfinity/dre/commit/1eda593b16fb3e72e6f102df708b5b69eb502bda#diff-13ee4b2252c9e516a0547f2891aa2105c3ca71c6d7a1e682c69be97998dfc87eL747)
Here is the fix we added yesterday:  [https://github.com/dfinity/dre/commit/f029720988584e4a24450be4f83c439c3bf4738e#diff-13ee4b2252c9e516a0547f2891aa[…]71c6d7a1e682c69be97998dfc87eL6037](https://github.com/dfinity/dre/commit/f029720988584e4a24450be4f83c439c3bf4738e#diff-13ee4b2252c9e516a0547f2891aa2105c3ca71c6d7a1e682c69be97998dfc87eL6037)
The issue is how the axum-otel-metrics works:
Axum otel metrics configures a global meter which is returned as a static variable. Since that global meter is a part of opentelemetry crate, if we have two different versions we won't see any of our custom metrics on service discovery.


sasa-tomic (approved):
TY!

PR #504, opened 2024-06-20
chore(dre): Rename --simulate to --dry-run

sasa-tomic:
- this makes it more consistent with ic-admin
- it's backwards compatible since the old arguments are still accepted


pietrodimarco-dfinity (approved):
Nice :)

LittleChimera (approved):

PR #528, opened 2024-06-25
Only include auth in ic-admin command lines when auth is in fact requ…

DFINITYManu:
…ired.

Solves problem with ic-admin get-subnet-list unexpectedly getting auth parameters when HSM_PIN variable is defined.


rs/cli/src/ic_admin.rs
```diff
@@ -382,7 +385,7 @@ impl IcAdminWrapper {
     }
 
     /// Run an `ic-admin get-*` command directly, and without an HSM
-    pub async fn run_passthrough_get(&self, args: &[String], silent: bool) -> anyhow::Result<String> {
+    pub async fn run_passthrough_get(&self, args: &[String], silent: bool, with_auth: bool) -> anyhow::Result<String> {
```
LittleChimera, line 388: get never requires auth. this argument will always be false

DFINITYManu:
Thanks    for the review.  It may be that I worked from an old  copy of main and this is fixed.  I am building and testing current head of main to see if it is fixed, as @sasa-tomic tells me it ought to be fixed.

PR #534, opened 2024-06-25
feat: enabling ci checks for release-index.yaml

NikolaMilosa:
Example of successful CI with warnings: https://github.com/dfinity/dre/actions/runs/9665911604/job/26664172436?pr=534
Example of CI with errors: https://github.com/dfinity/dre/actions/runs/9665980671/job/26664410826?pr=534

### DIFF
--- a/release-controller/ci_check.py
+++ b/release-controller/ci_check.py
@@ -32,7 +32,7 @@ def parse_args():
 
 
 def success_print(message: str):
-    print(f"{Fore.GREEN}{message}{Fore.RESET}")
+    print(f"{Fore.GREEN}✅ SUCCESS: {message}{Fore.RESET}")
 
 
 def print_and_ret(message: str) -> str:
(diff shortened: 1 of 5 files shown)


LittleChimera (approved):

release-controller/ci_check.py
```diff
@@ -35,66 +35,80 @@ def success_print(message: str):
...
-    print(f"{Fore.RED}{message}{Fore.RESET}")
+def print_and_ret(message: str) -> str:
+    print(message)
+    return message
 
 
-def warn_print(message: str):
-    print(f"{Fore.YELLOW}{message}{Fore.RESET}")
+def error_print(message: str) -> str:
+    return print_and_ret(f"{Fore.RED}{message}{Fore.RESET}")
```
sasa-tomic, line 44: I think it would be useful to print a prominent `ERROR:` or similar, to make it clear which line was an error. Color is great (let's keep it), but sometimes we're primed for letters, and may get confused about what the error actually was.

release-controller/ci_check.py
```diff
@@ -35,66 +35,80 @@ def success_print(message: str):
...
 
 
-def validate_schema(index: dict, schema_path: str):
+def warn_print(message: str) -> str:
+    return print_and_ret(f"{Fore.YELLOW}{message}{Fore.RESET}")
```
sasa-tomic, line 48: Similar thing here for warnings

sasa-tomic (approved):

release-controller/ci_check.py
NikolaMilosa: Added:
```
✅ SUCCESS: Schema validation passed
❌ ERROR: Schema validation failed
```

PR #535, opened 2024-06-26
fix(dre): asking for update everywhere

NikolaMilosa:
With this fix we will be able to:
1. Automatically reject upgrades if stdin is not a tty (useful if running in docker container)
  To test this you can run:
  ```bash
  echo "test" | dre registry
  ```
2. Set an environment variable to skip the check (useful if running in some scripts that are still run from terminal)
  To test this you can run:
  ```bash
  DRE_REFUSE_UPDATE=1 dre registry
  ```


DFINITYManu (approved):
This is good but it still won't fix the problem of running on a script in a terminal breaking if an update is published.

Please let's get that fixed too.

NikolaMilosa:
If you have it wrapped in a script, you can add the environment variable and it will skip the check. Why is that not enough?

PR #538, opened 2024-06-26
Refactor SubnetChangeRequest to work with Generics

pietrodimarco-dfinity:


sasa-tomic (approved):
Nice improvement, thanks!

PR #541, opened 2024-06-26
feat(dre): Speed up get operations by not preparing a neuron and the registry

sasa-tomic:
With this, running `dre get` now finishes in 0.2s.
Without the change, running the same command takes 13.3s.


pietrodimarco-dfinity (approved):
**NOICE**  :rocket:

PR #556, opened 2024-07-01
Moving out subnet management changes computation from Runner

pietrodimarco-dfinity:
The following:
1.  Adds `SubnetManager` struct independent from `Runner`
2. Moves subnet replacement computation in new `SubnetManager`
3. Adds RC registry as single point where to read registry from
4. Simplifies `membership_replace` logic


DFINITYManu (approved):

sasa-tomic (approved):
LGTM, thanks!

PR #557, opened 2024-07-01
refactor: renaming bazel targets

NikolaMilosa:
Convenience refactor for easier to use targets.
Previously to import a docker image into local registry we would have to run something like:
```
bazel run //rs/ic-observability/multiservice-discovery:multiservice-discovery-tarball
```
Now we can just run:
```
bazel run //rs/ic-observability/multiservice-discovery:tarball
```
And the target names are the same across all binaries.

Contains additional upgrade for pylint in pre-commit so that it works in envs with python3.12: 
Linked issue: https://github.com/pre-commit/pre-commit/issues/3041

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -67,7 +67,7 @@ repos:
           - --match=.*
 
   - repo: https://github.com/PyCQA/pylint
-    rev: v2.17.7
+    rev: v3.2.5
     hooks:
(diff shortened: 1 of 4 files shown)


sasa-tomic (approved):
TY!

DFINITYManu:
Late for the approval but thank you nonetheless.

PR #564, opened 2024-07-02
feat(dre): Show the date+time when voting, and do not stomp on previous logs

sasa-tomic:
I noticed that the first "vote" log line overwrites whatever the last log line. Fixed here.
Also, show the date & time of the voting in the logs, for easier auditing.

### DIFF
--- a/rs/cli/src/general.rs
+++ b/rs/cli/src/general.rs
@@ -102,7 +102,7 @@ pub async fn vote_on_proposals(
             let datetime = Local::now();
 
             info!(
-                "{} Voting on proposal {} (topic {:?}, proposer {}) -> {}\n",
+                "{} Voting on proposal {} (topic {:?}, proposer {}) -> {}",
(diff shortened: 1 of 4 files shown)


rs/cli/src/general.rs
```diff
@@ -97,8 +99,11 @@ pub async fn vote_on_proposals(
         print!("\x1B[1A\x1B[K");
         std::io::stdout().flush().unwrap();
         for proposal in proposals_to_vote.into_iter() {
+            let datetime = Local::now();
+
             info!(
-                "Voting on proposal {} (topic {:?}, proposer {}) -> {}",
+                "{} Voting on proposal {} (topic {:?}, proposer {}) -> {}\n",
```
sasa-tomic, line 105:
```suggestion
                "{} Voting on proposal {} (topic {:?}, proposer {}) -> {}",
```

PR #587, opened 2024-07-08
chore: updating rust toolchains

NikolaMilosa:


LittleChimera (approved):

sasa-tomic (approved):
TY!

rs/slack-notifications/src/main.rs
```diff
@@ -92,6 +92,7 @@ impl ProposalCheckpointStore {
         retry::retry(retry::delay::Exponential::from_millis(10).take(5), || {
             std::fs::OpenOptions::new()
                 .create(true)
+                .truncate(false)
```
sasa-tomic, line 95: Probably should be `truncate(true)` here.

rs/ic-observability/multiservice-discovery/src/definition.rs
```diff
@@ -534,6 +534,7 @@ impl DefinitionsSupervisor {
             retry::retry(retry::delay::Exponential::from_millis(10).take(5), || {
                 std::fs::OpenOptions::new()
                     .create(true)
+                    .truncate(false)
```
sasa-tomic, line 537: Probably should be `truncate(true)` here.

PR #622, opened 2024-07-16
ci: updating runner image to support cross compilation

NikolaMilosa:
Currently easiest way to support cross compilation is to use plain `cargo`.
Explored bazel possibilities like: 
* https://github.com/uber/hermetic_cc_toolchain
* https://github.com/bazelbuild/rules_rust/tree/main/examples/zig_cross_compiling

Right now both of the require to be run on actual `macos` since there is no support to add `--sysroot` for MacOS sdk.


pietrodimarco-dfinity (approved):
Thx!

PR #640, opened 2024-07-17
Add DRE tool command to query node-metrics-canister

pietrodimarco-dfinity:


rs/cli/src/commands/node_metrics/mod.rs
```diff
@@ -0,0 +1,42 @@
...
+#[derive(Args, Debug)]
+pub struct NodeMetricsCmd {
+    #[clap(subcommand)]
+    pub subcommand: NodeMetricsCommand,
+}
+
+#[derive(Subcommand, Debug)]
+pub enum NodeMetricsCommand {
+    FromSubnetManagementCanister(FromSubnetManagement),
+    FromNodeMetricsCanister(FromNodeMetrics),
```
NikolaMilosa, line 18: These would translate to `from-subnet-management-canister` and `from-node-metrics-canister`, maybe they are too verbose? If they have a common pattern (`from-...-canister`) then maybe that part is not needed?
What do you think about
```
dre node-metrics from management-canister
dre node-metrics from metrics-canister
```

Seems a bit more fluent? If you want to do something like that you can follow the example in `version` command. Maybe even `-canister` is too much since I would assume that all targets are canisters. Also would add some aliases like `mgmt` for `management` 😄

NikolaMilosa (approved):

rs/cli/src/commands/node_metrics/mod.rs
pietrodimarco-dfinity: I agree it would be more fluid but maybe not as meaningful as having `-canister` in it. but probably we can do the from subcommand ;)

sasa-tomic (approved):
Nice! TY!

PR #645, opened 2024-07-18
feat(dre): adding an option for dre to not sync with the nns

NikolaMilosa:
Useful if the NNS is down to allow at least some functionality of the tool.


sasa-tomic (approved):
nice! thank you!

PR #648, opened 2024-07-19
Split IcAdminRequirement in AuthRequirement and NeuronRequirements

pietrodimarco-dfinity:


DFINITYManu (approved):
beautiful.  i was looking forward to this specific development.

rs/cli/src/auth.rs
```diff
@@ -18,93 +18,43 @@ use log::info;
...
 #[derive(Clone, Debug)]
 pub struct Neuron {
-    pub auth: Auth,
     pub neuron_id: u64,
     pub include_proposer: bool,
 }
 
+#[allow(unused)]
 static RELEASE_AUTOMATION_DEFAULT_PRIVATE_KEY_PEM: &str = ".config/dfx/identity/release-automation/identity.pem"; // Relative to the home directory
+#[allow(unused)]
```
LittleChimera, line 27: if it's not used, remove it

rs/cli/src/commands/mod.rs
```diff
@@ -236,11 +236,44 @@ pub trait ExecutableCommand {
...
 
-pub enum IcAdminRequirement {
-    None,
-    Anonymous,                                          // for get commands
-    Detect,                                             // detect the neuron
-    OverridableBy { network: Network, neuron: Neuron }, // eg automation which we know where is placed
+pub enum NeuronRequirement {
+    None,       // for get commands
+    AutoDetect, // detect the neuron
+    #[allow(unused)]
```
LittleChimera, line 242: ditto

LittleChimera (approved):

rs/cli/src/auth.rs
```diff
@@ -156,10 +106,60 @@ impl Auth {
...
+            maybe_neuron: None,
+        }
+    }
+
+    pub fn with_neuron(self, neuron: Neuron) -> Self {
+        Self {
+            maybe_neuron: Some(neuron),
+            ..self
+        }
+    }
```
NikolaMilosa, line 130: Kind of against allowing this since this means we can always override it after initialization. The idea for safe execution would be to initialize within `ctx.rs` and after that you can only use what you configured. This means that magically someone from somewhere may mutate this specific thing is what I tried to avoid. After the call to construct the `DreContext` one should only use what he configured. No mutations or translations. This feels like a hack

rs/cli/src/auth.rs
```diff
@@ -156,10 +106,60 @@ impl Auth {
...
+        let neuron = Neuron {
+            neuron_id: self.auto_detect_neuron_id(&network.nns_urls).await?,
+            include_proposer: true,
+        };
+
+        Ok(Self {
+            maybe_neuron: Some(neuron),
+            ..self
+        })
+    }
```
NikolaMilosa, line 142: Similar as above. Better to provide one constructor and disallow the calls that will take ownership and transform. If there is a quick fix people will use it

rs/cli/src/commands/api_boundary_nodes/add.rs
```diff
@@ -23,7 +23,7 @@ pub struct Add {
 
 impl ExecutableCommand for Add {
     fn require_ic_admin(&self) -> IcAdminRequirement {
-        IcAdminRequirement::Detect
+        IcAdminRequirement::default()
```
NikolaMilosa, line 26: What is default? This means I have to go the specific code to see which thing is default. Also the default can change and all the commands would change the requirement. I would prefer being explicit rather than having language magic for configuration of commands

rs/cli/src/commands/completions.rs
```diff
@@ -11,7 +11,7 @@ pub struct Completions {
 
 impl ExecutableCommand for Completions {
     fn require_ic_admin(&self) -> super::IcAdminRequirement {
-        super::IcAdminRequirement::None
+        super::IcAdminRequirement::none()
```
NikolaMilosa, line 14: If you want to, you can name them and return via functions just please don't use `default`.

NikolaMilosa (commented):
Don't know, feels like we moved a lot of code just to tackle a one-off scenario that is super specific. Having a `neuron_id` on 0 for calling non governance canisters sounds okay to me. Plus we lost the possibility to override which was super useful for when you forget to use a different neuron for certain proposals like manual placing of version upgrade proposal :/

pietrodimarco-dfinity:
> Don't know, feels like we moved a lot of code just to tackle a one-off scenario that is super specific. Having a `neuron_id` on 0 for calling non governance canisters sounds okay to me. Plus we lost the possibility to override which was super useful for when you forget to use a different neuron for certain proposals like manual placing of version upgrade proposal :/

Ok lets not do it then, will keep the branch if it will become more useful having something like this in the future and we can adapt the code

PR #684, opened 2024-08-02
Add node rewards computation in backend canister

pietrodimarco-dfinity:


rs/dre-canisters/trustworthy-node-metrics/src/trustworthy-node-metrics/src/rewards_manager.rs
```diff
@@ -0,0 +1,101 @@
+use chrono::{DateTime, Duration, Utc};
+
+use crate::types::{NodeMetrics, Rewards, TimestampNanos};
+
+#[derive(Debug)]
+pub struct DailyFailureRate {
```
sasa-tomic, line 6: How about naming it to something more specific, e.g. `ProposeBlockFailureRateDaily`?

PR #696, opened 2024-08-08
chore: Bump a few deps

sasa-tomic:


Cargo.toml
```diff
@@ -77,7 +77,7 @@ custom_error = "1.9.2"
 decentralization = { path = "rs/decentralization" }
 trustworthy-node-metrics = { path = "rs/dre-canisters/trustworthy-node-metrics/src/trustworthy-node-metrics" }
 derive_builder = "0.20.0"
-derive_more = "0.99.18"
+derive_more = "*"
```
NikolaMilosa, line 80: Do we want to pick the latest compatible version always?

NikolaMilosa (approved):
Looks okay, a lot of versions changed, could be that we find something along the way

Cargo.toml
sasa-tomic: You're right, let me nail it to version 1 (latest)

PR #706, opened 2024-08-10
feat: autoupdate ic deps

NikolaMilosa:
A small script that will be run as a scheduled job to create a PR with new IC versions.


.github/workflows/update-dependencies.yaml
```diff
@@ -4,7 +4,7 @@ on:
   workflow_dispatch:
   schedule:
     # * is a special character in YAML so you have to quote this string
```
LittleChimera, line 6:
```suggestion
```

ic-revisions.json
```diff
@@ -0,0 +1,10 @@
+{
+	"https://github.com/dfinity/ic.git": {
+		"commit": "7dee90107a88b836fc72e78993913988f4f73ca2",
+		"ref": "refs/heads/master" 
+	},
+	"https://github.com/dfinity/cdk-rs.git": {
+		"commit": "59795716487fbb8a9910ac503bcea1e0cb08c932",
+		"ref": "refs/heads/main"
+	}
+}
```
LittleChimera, line 10:
```suggestion
}

```

NikolaMilosa: This file is managed by the script, I don't think that its necessary to format it :D

NikolaMilosa: What i mean is that it will automatically have a new line

LittleChimera (approved):

PR #712, opened 2024-08-13
Improving Node-Metrics backend canister

pietrodimarco-dfinity:
- Decreased number of instruction needed to update metrics.
This is especially relevant when stable memory is empty (before it risked exhausting the number of instructions allowed for timers) 37B before vs  700M now.
 This was mainly due to the multiple accesses to stable map using `metrics_before_ts` querying the metrics before a ts for a specific `node_id`. Replaced now with `latest_metrics` which goes through the data only once. 

The use of unbounded data for `NodeMetricsStored`

- Added Virtual Memory needed for storing additional data structures (node providers mapping in the future)

- Modified `NodeMetricsStored` serialization to be bounded


sasa-tomic (approved):
:tada:

PR #729, opened 2024-08-15
Do not alert for proposals already voted by Neuron 27 (Dfinity neuron)

pietrodimarco-dfinity:
- Add governance canister client to interact with gov. canister
- Uses `get_neuron_info` for getting most recent (100) proposals voted on by the neuron
- Moved alert formatting in `alerts_receiver`
- Added tests for classes


release-controller/release_index.py
```diff
@@ -26,12 +24,12 @@ class Release(BaseModel):
...
     model_config = ConfigDict(
         extra='forbid',
     )
     releases: List[Release]
 
 
-class Model(RootModel[ReleaseIndex]):
-    root: ReleaseIndex
+class Model(RootModel[Welcome4]):
+    root: Welcome4
```
NikolaMilosa, line 35: Was this renaming intentional?

trusted-neurons-alerts/governance.py
```diff
@@ -0,0 +1,79 @@
...
+            version = self._version()
+            governance_did = pathlib.Path(tmpdirname) / "governance.did"
+            contents = requests.get(
+                f"https://raw.githubusercontent.com/dfinity/ic/{version}/rs/nns/governance/canister/governance.did",
+                timeout=10,
+            ).text
+            with open(governance_did, "w", encoding="utf8") as f:
+                f.write(contents)
+
+            governance_canister = Canister(agent=self.agent, canister_id=self.principal, candid=open(governance_did, encoding="utf8").read())
```
NikolaMilosa, line 51: Not sure if this is supported in `ic-py` but I know there is a way to fetch candid directly from a canister via metadata. It would remove the dependency to github and allow you to always be sure to have a correct version of candid

NikolaMilosa (approved):

trusted-neurons-alerts/governance.py
pietrodimarco-dfinity: This would be cool! I will look for it

release-controller/release_index.py
pietrodimarco-dfinity: nope let me revert

PR #730, opened 2024-08-15
Do not alert for proposals already voted by Neuron 27 (Dfinity neuron)

pietrodimarco-dfinity:
- Add governance canister client to interact with gov. canister
- Uses `get_neuron_info` for getting most recent (100) proposals voted on by the neuron
- Moved alert formatting in `alerts_receiver`
- Added tests for classes


sasa-tomic (approved):
Nice!

PR #733, opened 2024-08-15
chore: Update dependencies

sa-github-api:
This PR updates poetry (Python) and cargo (Rust) dependencies and IC repository dependencies

### DIFF
--- a/release-controller/release_index.py
+++ b/release-controller/release_index.py
@@ -3,7 +3,7 @@
 
 from __future__ import annotations
 
-from typing import List
+from typing import List, Optional
 
 from pydantic import BaseModel, ConfigDict, RootModel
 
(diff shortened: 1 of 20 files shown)


release-controller/release_index.py
```diff
@@ -26,12 +24,12 @@ class Release(BaseModel):
     versions: List[Version]
 
 
-class ReleaseIndex(BaseModel):
+class Welcome4(BaseModel):
```
LittleChimera, line 27: what happened here?

NikolaMilosa: Apparently we have a `/bin/poetry.sh` which has `datamodel-codegen` which uses `release-index-schema.json` and generates that file. I will make changes in a separate PR to change the root node name to fix this

NikolaMilosa (approved):

PR #734, opened 2024-08-16
fix(dre): [DRE-241] Prefer existing nodes in subnet when selecting best results

sasa-tomic:
This commit modifies the `DecentralizedSubnet` implementation in the `network.rs` file to prefere existing nodes in the subnet when selecting the best results.

This change ensures that the same nodes are retained in the subnet if they are already there.

### DIFF
--- a/rs/decentralization/src/network.rs
+++ b/rs/decentralization/src/network.rs
@@ -403,6 +403,16 @@ impl DecentralizedSubnet {
         if best_results.is_empty() {
             None
         } else {
+            // If any of the best_results nodes are already in the subnet,
+            // we should prefer them. This is because we want to keep the
+            // same nodes in the subnet if they are already there.
+            let current_nodes_set: HashSet<_> = current_nodes.iter().collect();
+            for result in best_results {
+                if current_nodes_set.contains(&result.node) {
+                    return Some(result.clone());
+                }
+            }
+
             // We sort the current nodes by alphabetical order on their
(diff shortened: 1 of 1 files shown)


NikolaMilosa (approved):

LittleChimera (approved):

rs/decentralization/src/network.rs
```diff
@@ -3,6 +3,7 @@ use crate::subnets::unhealthy_with_nodes;
 use crate::SubnetChangeResponse;
 use actix_web::http::StatusCode;
 use actix_web::{HttpResponse, ResponseError};
+use ahash::{AHashSet, HashSet};
```
sasa-tomic, line 6:
```suggestion
use ahash::AHashSet;
```

rs/decentralization/src/network.rs
```diff
@@ -403,6 +404,16 @@ impl DecentralizedSubnet {
         if best_results.is_empty() {
             None
         } else {
+            // If any of the best_results nodes are already in the subnet,
+            // we should prefer them. This is because we want to keep the
+            // same nodes in the subnet if they are already there.
+            let current_nodes_set: HashSet<_> = current_nodes.iter().collect();
```
sasa-tomic, line 410:
```suggestion
            let current_nodes_set: AHashSet<_> = current_nodes.iter().collect();
```

PR #736, opened 2024-08-16
feat(dre): [DRE-240] Add more info to the subnet membership change proposals

sasa-tomic:
This should make it easier to understand the motivation behind the proposal, and to verify the proposal payload.


pietrodimarco-dfinity (approved):
Very cool!

PR #742, opened 2024-08-16
feat(release-controller): various improvements

LittleChimera:
* improved testing - enabled testing of individual commits and integrated bazel into bazel
* improved detecting to rely on just files instead of packages
* detecting versions bumps
* publishing markdown with exclusions to dre repo (but electing without it still)
* add a reason for exclusion


NikolaMilosa (approved):
Slapping a fat 💯

PR #743, opened 2024-08-16
feat(qualification): implementing multiple starting versions

NikolaMilosa:
Since it is possible to have multiple versions forcasted for the end of an active rollout this PR addresses qualification to all those versions. It adds modes to the qualifier which both have their pros and cons explained in the code.

Even with this we might be missing some scenario:

Say that not all 40 subnets are rolled out during this active rollout. This means that some subnets will stay on their current version. This will be addressed in a separate PR.


rs/qualifier/src/main.rs
```diff
@@ -33,26 +39,69 @@ async fn main() -> anyhow::Result<()> {
...
-        info!("Fetching the forcasted version of NNS which will be used as starting point");
+        info!("Fetching the forcasted versions from mainnet which will be used as starting point");
         // Fetch the starter versions
         let start_version_selector = StartVersionSelectorBuilder::new()
             .with_client(ClientBuilder::new().connect_timeout(Duration::from_secs(30)))
             .build()
             .await?;
 
-        start_version_selector.get_forcasted_version_for_mainnet_nns()?
+        &start_version_selector.get_forcasted_versions_from_mainnet()?
```
sasa-tomic, line 53: https://word.tips/spelling/forecast-vs-forcast/ :innocent:

sasa-tomic (approved):

PR #745, opened 2024-08-16
Modify rewards computation on Node-Rewards canister

pietrodimarco-dfinity:
As result with meeting with @bjoernek on node rewards.

The PR:
- Modifies node rewards computation by applying linear reduction on the avg. failure rate on the entire reward period
- Removed node rewards with penalty
- Update tests for node rewards computation 
- Add stats for node rewards computation


sasa-tomic (approved):
Nice, thank you!

PR #753, opened 2024-08-19
update dev setup for python

LittleChimera:


sasa-tomic (approved):
LGTM

docs/contributing.md
```diff
@@ -82,11 +82,10 @@ Run the following from the repo root:
 ```bash
 # change into the directory of the repo
 # cd ~/src/release
-pyenv install 3.11.6  # installs Python 3.11.6 via pyenv
-pyenv local 3.11.6    # tells pyenv to use 3.11.6 for this repo
-pip3 install poetry   # installs poetry to your 3.11.6
-poetry env use $(which python)  # instructs poetry to use 3.11.6
-poetry install        # installs all our dependencies to 3.11.6
+pyenv install # installs Python from .python-version
```
sasa-tomic, line 85: Maybe would be better to install python 3.11 instead of a random (latest) python version?

PR #758, opened 2024-08-20
chore: add alias for proposals -> proposal subcommand

sasa-tomic:
In some subcommands (e.g. "get") we work work with a single proposal, so a singular "proposal" makes more sense. But changing to singular wouldn't make sense for some other subcommands. So let's have both!


NikolaMilosa (approved):
Can't wait to do a 
```bash
dre proposal filter --limit 150
```

PR #759, opened 2024-08-20
Add node providers in Node Rewards canister

pietrodimarco-dfinity:
**Changes:**
- Adds node providers fetching and storage per node
- Modifies `NodeRewardsResponse` to return the `node_provider_id` per node


sasa-tomic (approved):
Lgtm!

PR #844, opened 2024-09-02
Refactor authentication detection.

DFINITYManu:
The fundamental change in this PR could have been described as reducing ambiguity in command line processing for the authentication options, by exploiting the Rust type system.  The path was somewhat more tortuous.

Chiefly, we simplify a lot of the inner function calls, which no longer require four different arguments that could each wrongly and independently have missing values when they should never have them.  Argument combinations that are invalid must be rejected.

* Authentication selection:
  * Use of `--private-key-pem` with any HSM-related argument is forbidden.
  * Hardware security module detection now allows the user to narrow down to either an HSM slot, or an HSM key ID, or both, or let autodetection take place as usual.
    * When none are specified, autodetection remains unchanged from before (although arguably when autodetection fails, the program should abort instead of falling back to `::Anonymous` -- that is, if the user's desired command can work with anonymous authentication, autodetection shouldn't even take place to begin with).
    * When an HSM slot is specified, but an HSM is not detected in that slot, the program aborts.
    * When an HSM key ID is specified, but the key ID is not detected in any slot (or the specified slot), the program aborts.
  * All failure / error messages are clear and explain to the user what the problem is.  We no longer bark "Invalid auth parameters" to the user.
* Key ID fixes:
  * The key ID was used both when the DRE tool invokes `ic-admin` via an `exec()` call, and when a canister client is instantiated.
  * Under autodetection, DRE was always hardcoding `01` as the key ID, which is wrong (it only worked by coincidence).
  * We now *detect the correct key ID* (basically, the first object that is a `Token` in the slot) and use that value for `ic-admin`.
  * Key IDs are technically large integers (Vec<u8>) but for simplicity we only support up to 256 key IDs, period.  The text format that the canister client wants (a `String` of hex characters for some odd reason, instead of a `Vec<u8>` or a simple `u64`) is respected.
  * As a result of the key ID change, it is no longer possible for users to supply an invalid key ID.  It continues to be possible for users to supply a nonexistent key ID, but we check it's there before proceeding with it.
* PIN entry fixes:
  * PIN entry is saved to the keyring **only** if the PIN is valid.  The PIN is first tested before saving and using it, and if the PIN is bad, the program aborts and explains why.  The keyring is always cleared upon invalid PIN, to prevent another re-run with a bad PIN, and therefore protect against the exhaustion of PIN entry tries in the HSM, and therefore protect against the HSM being locked (guess whom it happened to first, lol).
  * PIN is never saved in the keyring if specified via `--hsm-pin` or the `HSM_PIN` environment variable.
  * The program warns users when their HSM is about to run out of PIN entry tries, *before* entering the PIN.
  * The program aborts if an attempt to use a locked HSM is made.  The user is informed appropriately.
    * In the unlikely case that the user has two HSMs plugged in and autodetection selected one that just happened to be locked, autodetection fails, but the user can still specify another slot using the `--hsm-slot` argument.

We now have a luxury authentication detection system for DRE!

Additional features:

* The voting loop can now be limited to one cycle, by setting the --sleep-time to zero seconds.
* The flag `--dry-run` is now respected for `dre vote`.
* Some functions that used blocking calls have been converted to `async`; the blocking sections of these functions are spawned using `tokio::blocking_spawn` to prevent starvation.

### DIFF
--- a/rs/cli/src/auth.rs
+++ b/rs/cli/src/auth.rs
@@ -31,6 +31,16 @@ pub struct Neuron {
 static RELEASE_AUTOMATION_DEFAULT_PRIVATE_KEY_PEM: &str = ".config/dfx/identity/release-automation/identity.pem"; // Relative to the home directory
 const RELEASE_AUTOMATION_NEURON_ID: u64 = 80;
 
+// As per fn str_to_key_id(s: &str) in ic-canisters/.../parallel_hardware_identity.rs,
+// the representation of key ID that the canister client wants is a sequence of
+// pairs of hex digits, case-insensitive.  The key ID as stored in the HSM is
+// a Vec<u8>.  We only store the little-endianest of the digits from that Vec<> in
+// our key_id variable.  The following function produces what the ic-canisters
+// code wants.
+pub fn hsm_key_id_to_string(s: u8) -> String {
+    format!("{:02x?}", s)
+}
+
 impl Neuron {
     pub async fn new(auth: Auth, neuron_id: Option<u64>, network: &Network, include_proposer: bool) -> anyhow::Result<Self> {
         let neuron_id = match neuron_id {
@@ -116,20 +126,10 @@ impl Auth {
     }
 
     async fn auto_detect_neuron_id(&self, nns_urls: &[url::Url]) -> anyhow::Result<u64> {
-        // As per fn str_to_key_id(s: &str) in ic-canisters/.../parallel_hardware_identity.rs,
-        // the representation of key ID that the canister client wants is a sequence of
-        // pairs of hex digits, case-insensitive.  The key ID as stored in the HSM is
-        // a Vec<u8>.  We only store the little-endianest of the digits from that Vec<> in
-        // our key_id variable.  The following function produces what the ic-canisters
-        // code wants.
-        fn key_id_to_str(s: u8) -> String {
-            format!("{:02x?}", s)
-        }
-
         // FIXME: why do we even take multiple URLs if only the first one is ever used?
         let url = nns_urls.first().ok_or(anyhow::anyhow!("No NNS URLs provided"))?.to_owned();
         let client = match self {
-            Auth::Hsm { pin, slot, key_id } => IcAgentCanisterClient::from_hsm(pin.clone(), *slot, key_id_to_str(*key_id), url, None)?,
+            Auth::Hsm { pin, slot, key_id } => IcAgentCanisterClient::from_hsm(pin.clone(), *slot, hsm_key_id_to_string(*key_id), url, None)?,
             Auth::Keyfile { path } => IcAgentCanisterClient::from_key_file(path.clone(), url)?,
             Auth::Anonymous => IcAgentCanisterClient::from_anonymous(url)?,
         };
--- a/rs/cli/src/auth.rs
+++ b/rs/cli/src/auth.rs
@@ -116,7 +116,8 @@ impl Auth {
                 // and must fail upfront if autodetection fails, and other operations do not require
                 // authentication and should not even attempt to perform HSM autodetection.
                 // Ideally that means the operations that require authentication, and only those
-                // operations, accept the command line arguments that create authentication.
+                // operations, accept the command line arguments that create authentication --
+                // other operations should outright reject the arguments.
                 (None, None, None) => Ok(Self::detect_hsm_auth()?.map_or(Auth::Anonymous, |a| a)),
                 _ => Err(anyhow::anyhow!("When specifying either --hsm-slot, --hsm-pin or --hsm-key-id, autodetection is disabled and all three arguments must be specified")),
             },
--- a/rs/cli/src/auth.rs
+++ b/rs/cli/src/auth.rs
@@ -1,19 +1,25 @@
 use std::path::PathBuf;
 use std::str::FromStr;
 
+use anyhow::anyhow;
 use clio::InputPath;
+use cryptoki::object::AttributeInfo;
+use cryptoki::session::Session;
 use cryptoki::{
     context::{CInitializeArgs, Pkcs11},
-    session::{SessionFlags, UserType},
+    object::{Attribute, AttributeType},
+    session::UserType,
+    slot::{Slot, TokenInfo},
 };
 use dialoguer::{console::Term, theme::ColorfulTheme, Password, Select};
 use ic_canisters::governance::GovernanceCanisterWrapper;
 use ic_canisters::IcAgentCanisterClient;
 use ic_management_types::Network;
 use keyring::{Entry, Error};
-use log::info;
+use log::{info, warn};
+use secrecy::SecretString;
 
-use crate::commands::{AuthOpts, HsmOpts};
+use crate::commands::{AuthOpts, HsmOpts, HsmParams};
 
 #[derive(Clone, Debug)]
 pub struct Neuron {
@@ -31,6 +37,7 @@ impl Neuron {
             Some(n) => n,
             None => auth.auto_detect_neuron_id(network.get_nns_urls()).await?,
         };
+        info!("Identifying as neuron ID {}", neuron_id);
         Ok(Self {
             auth,
             neuron_id,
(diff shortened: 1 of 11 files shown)


rs/cli/src/auth.rs
```diff
@@ -220,3 +199,28 @@ impl Auth {
...
+                private_key_pem: _,
+                hsm_opts: Some(hsm_opts),
+            } => Ok(Auth::Hsm {
+                pin: hsm_opts.hsm_pin.clone(),
+                slot: hsm_opts.hsm_slot,
+                key_id: hsm_opts.hsm_key_id.clone(),
+            }),
+            // I think the next line should not fall back to anonymous.
+            // It should always be autodetect and fail if detection fails.
+            _ => Ok(Self::detect_hsm_auth()?.map_or(Auth::Anonymous, |a| a)),
```
NikolaMilosa, line 223: Not really, because if you want to do some canister interactions, you don't need ic-admin but you need some auth, usually a readonly auth, that is why we have `Anonymous` here. But sometimes you do need some auth for canisters

rs/cli/src/commands/mod.rs
```diff
@@ -48,24 +51,90 @@ pub mod upgrade;
...
+    )]
+    pub hsm_key_id: String,
+}
 
-    /// Path to key pem file
-    #[clap(long, global = true, env = "PRIVATE_KEY_PEM")]
-    pub private_key_pem: Option<PathBuf>,
+// Stupid clap does not let me define this
+// and the above struct as enums.
+// https://github.com/clap-rs/clap/issues/2621
```
NikolaMilosa, line 94: Understand the frustration, but maybe not leave it like this in a public repo 😟

NikolaMilosa (commented):
Looks like a lot of dependencies changed somehow. Can you double check?

rs/cli/src/commands/mod.rs
DFINITYManu: Fair.

DFINITYManu: This has been addressed.  PTAL @NikolaMilosa .

rs/cli/src/auth.rs
DFINITYManu: While I won't change this here (I'm leaving it as-is), this definitely needs to be fixed.  In those "some interactions that don't need ic-admin but some auth", a different **type** should be used.  And the program does know this already -- but the circumstance is just not specified in types as it should be.  It should be impossible -- enforced by the type system -- for an action that requires full auth to ever execute with anonymous auth.

rs/cli/src/auth.rs
```diff
@@ -22,21 +32,23 @@ pub struct Neuron {
...
-        include_proposer: bool,
-    ) -> anyhow::Result<Self> {
-        let auth = Auth::from_cli_args(private_key_pem, hsm_slot, hsm_pin, hsm_key_id)?;
+    pub async fn new(auth: Auth, neuron_id: Option<u64>, network: &Network, include_proposer: bool) -> anyhow::Result<Self> {
         let neuron_id = match neuron_id {
             Some(n) => n,
-            None => auth.auto_detect_neuron_id(network.get_nns_urls()).await?,
+            None => auth.auto_detect_neuron_id(network.get_nns_urls().to_vec()).await?,
         };
+        info!("Identifying as neuron ID {}", neuron_id);
```
NikolaMilosa, line 51: I think our tool is becoming way too verbose, I suggest change  this to debug/trace

rs/cli/src/auth.rs
```diff
@@ -55,7 +67,10 @@ impl Neuron {
         vec![]
     }
 
+    // FIXME: there should be no unchecked anything.
+    // Caller must be able to bubble up the error of the file not existing there.
     pub fn automation_neuron_unchecked() -> Self {
+        info!("Constructing neuron using the release automation private key");
```
NikolaMilosa, line 73: Same as above

rs/cli/src/auth.rs
```diff
@@ -66,6 +81,7 @@ impl Neuron {
     }
 
     pub fn anonymous_neuron() -> Self {
+        info!("Constructing anonymous neuron (ID 0)");
```
NikolaMilosa, line 84: Same for most of the `info` messages

rs/cli/src/auth.rs
```diff
@@ -125,60 +126,26 @@ impl Auth {
...
-                info!("HSM login successful!");
-                pin_entry.set_password(&pin)?;
-                return Ok(Some(Auth::Hsm {
-                    pin,
-                    slot: slot.id(),
-                    key_id: "01".to_string(),
-                }));
-            }
+    pub fn create_canister_client(&self, nns_urls: Vec<url::Url>, lock: Option<Mutex<()>>) -> anyhow::Result<IcAgentCanisterClient> {
+        // FIXME: why do we even take multiple URLs if only the first one is ever used?
```
NikolaMilosa, line 130: Because how you should implement a real `Agent` is to have a pool of agents, and have fallbacks if the node you are using is unhealthy. Generally if its https://ic0.app, boundary nodes do the load balancing for you, but if its a specific node ipv6 then you are on your own.

rs/cli/src/ctx.rs
```diff
@@ -39,67 +40,86 @@ pub struct DreContext {
     skip_sync: bool,
     ic_admin_path: Option<String>,
     forum_post_link: Option<String>,
+    dry_run: bool,
```
NikolaMilosa, line 43: This is a strictly ic-admin thing. Not sure if its needed here?

rs/cli/src/auth.rs
DFINITYManu: We should try all URLs when there's a failure in creating the canister client with the first URL.

Or actually the canister client itself should do that, like etcd or Zookeeper clients do.

rs/cli/src/ctx.rs
DFINITYManu: Dry run for other commands, like `dre vote`.

Of course I would prefer if the `--dry-run` flag was (by type) limited to only the commands that support it, but currently it is a global flag.  But that's a bigger refactor.

rs/cli/src/auth.rs
NikolaMilosa: The real answer is that `ic-admin` is just a wrapper for some canister calls. Meaning that, in theory we don't really need `ic-admin` and that is why we have the difference. Under the hood ic admin has a `ic-canister-client` which has a pool of agents.

rs/cli/src/auth.rs
NikolaMilosa: Both are correct, I would rather if agent has that logic it self, but I am not 100% sure if it does

rs/cli/src/ctx.rs
NikolaMilosa: Ok, thanks for clarification

rs/cli/src/commands/mod.rs
```diff
@@ -48,24 +50,71 @@ pub mod upgrade;
...
-    pub hsm_slot: Option<u64>,
+    #[clap(required = false,
+        conflicts_with = "private_key_pem",
+        long, value_parser=maybe_hex::<u64>, global = true, env = "HSM_SLOT")]
+    pub(crate) hsm_slot: Option<u64>,
 
     /// HSM Key ID, can be read with pkcs11-tool
-    #[clap(long, global = true, env = "HSM_KEY_ID")]
-    pub hsm_key_id: Option<String>,
+    #[clap(required = false, conflicts_with = "private_key_pem", long, value_parser=maybe_hex::<u8>, global = true, env = "HSM_KEY_ID")]
```
sasa-tomic, line 63: What's the use case for the `conflicts_with`?

DFINITYManu:
I am just happy I can now vote using `dre`.

sasa-tomic (approved):

rs/cli/src/commands/mod.rs
DFINITYManu: It rejects the use of the flag if the other flag is present.

rs/cli/src/commands/mod.rs
```diff
@@ -48,24 +50,71 @@ pub mod upgrade;
...
+#[derive(ClapArgs, Debug, Clone)]
+#[group(multiple = false)]
+/// Authentication arguments
+pub(crate) struct AuthOpts {
+    /// Path to private key file (in PEM format)
+    #[clap(
+        long,
+        required = false,
+        global = true,
+        conflicts_with_all = ["hsm_pin", "hsm_slot", "hsm_key_id"],
```
sasa-tomic, line 106: Same question here - is the restriction needed?

rs/cli/src/commands/mod.rs
DFINITYManu: Technically we could use groups for this, but groups aren't working at all here, for some reason.

PR #854, opened 2024-09-03
fix: registry canister

NikolaMilosa:
Migrating to one used in `rs-agent`

Fixes:
1. `dre registry`
2. `multiservice discovery`


sasa-tomic (approved):
:rocket: TY!

PR #857, opened 2024-09-03
feat: Replace poetry with rye for managing Python dependencies

sasa-tomic:
Info on rye can be found at https://rye.astral.sh/


bin/release-controller-update-data-model.sh
```diff
@@ -0,0 +1,14 @@
...
+set -eEuo pipefail
+
+git_root=$(git rev-parse --show-toplevel)
+cd "$git_root"
+command -v rye >/dev/null \
+    || rye run datamodel-codegen \
+        --input $git_root/release-index-schema.json \
+        --input-file-type jsonschema \
+        --output $git_root/release-controller/release_index.py \
+        --target-python-version 3.10 \
```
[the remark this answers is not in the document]
LittleChimera: or maybe parse it from somewhere

docker/Dockerfile
```diff
@@ -114,7 +60,7 @@ ENV RUSTUP_HOME=/opt/rustup \
     PATH=/opt/cargo/bin:$PATH
 
 # Make sure /rust-toolchain.toml has the same version
-ARG rust_version=1.66.1
+ARG rust_version=1.79.0
```
LittleChimera, line 63: could we get this from rust config instead?

sasa-tomic: yep!

PR #862, opened 2024-09-04
Address various issues through the use of Cargo deny.

DFINITYManu:
* Licensing compliance enforcement (needs to be reviewed).
* Banning the use of OpenSSL altogether, to prevent problems with dynamic libraries.

CI/CD still needs to be addressed here.

Also, there is a package that is GPL 3.0 and therefore we cannot use it.  This package is pulled in by `chrome_headless` and therefore we should remove that (to my understanding the screenshotting functionality doesn't currently work anyway).  If we really want that, we should consider rearchitecting that functionality to be a separate program that gets invoked via `execve()` from the qualifier and takes the screenshots autonomously during qualification; that way, that program does not force the entire code base to link with the GPL library `chrome_headless` includes, which is incompatible with the qualifier code.  We will then obviously have to add a screenshot-program-specific exception for that one in `deny.toml`, which is a nice self-referential reference, so that we can deny while we deny.


Cargo.lock
```diff
@@ -6475,50 +6427,12 @@ version = "0.3.1"
...
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a948666b637a0f465e8564c73e89d4dde00d72d4d473cc972f390fc3dcee7d9c"
-dependencies = [
- "proc-macro2",
- "quote",
- "syn 2.0.76",
-]
-
 [[package]]
 name = "openssl-probe"
```
LittleChimera, line 6431: hmmm, i also removed this one, but not sure why we still have it here.

NikolaMilosa (approved):

Cargo.lock
DFINITYManu: Due to the work that @sasa-tomic put into updating deps, a ton of the lock file stuff went away, and also a ton of the duplicate package exceptions we had in `deny.toml` was no longer necessary.

PR #863, opened 2024-09-04
chore: Remove some obsolete mentions of gitlab

sasa-tomic:


.github/workflows/main.yaml
```diff
@@ -109,7 +109,6 @@ jobs:
             bases/apps/mainnet-dashboard/backend/base/deployment.yaml
             bases/apps/mainnet-dashboard/statefulset-slack.yaml
             bases/apps/service-discovery/service-discovery.yaml
-            gitlab-ci/vector-configs.yml
```
NikolaMilosa, line 112: Instead of removing it lets just change the path to: `.github/workflows/dre-vector-configs.yaml`
Source: https://github.com/dfinity-ops/k8s/blob/main/.github/workflows/dre-vector-configs.yaml

NikolaMilosa:
This is fixed here: https://github.com/dfinity/dre/pull/865#pullrequestreview-2282521063 by @LittleChimera

PR #864, opened 2024-09-04
feat: overriding ic admin versions

NikolaMilosa:
Adds possibility to specify a version of `ic-admin` to wrap.

Previously we were forced to rely on governance canister version and fallback to a default.
Now we have the full control. The default is `from-governance` which is the same behavior as before.

To run dre with `ic-admin` with governance version:
```bash
dre --ic-admin-version from-governance/governance/govn/g get subnet-list
```

To run dre with `ic-admin` from default version that is updated when we update deps:
```bash
dre --ic-admin-version default/d get subnet-list
```

To run dre with `ic-admin` from a specific commit we can use:
```bash
dre --ic-admin-version <commit> get subnet-list
```

### DIFF
--- a/rs/cli/src/commands/mod.rs
+++ b/rs/cli/src/commands/mod.rs
@@ -20,6 +20,7 @@ use proposals::Proposals;
 use propose::Propose;
 use qualify::Qualify;
 use registry::Registry;
+use strum::Display;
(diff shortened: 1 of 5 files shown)


sasa-tomic (approved):
LGTM

DFINITYManu (approved):

PR #870, opened 2024-09-05
fix: removing duplicated rye and sourcing it in `update dependencies` job 

NikolaMilosa:


.github/workflows/update-dependencies.yaml
```diff
@@ -23,7 +23,9 @@ jobs:
       - name: "🐍 Install rye"
```
DFINITYManu, line 23: why a duck

DFINITYManu (approved):

DFINITYManu:
I will reapprove when fix is gud.

.github/workflows/update-dependencies.yaml
```diff
@@ -23,7 +23,9 @@ jobs:
...
         run: |
           export RYE_INSTALL_OPTION="--yes"
+          export RYE_HOME="$HOME/.rye"
           curl -sSf https://rye.astral.sh/get | bash
```
sasa-tomic, line 27: Why not have it in container only?

PR #872, opened 2024-09-05
fix: fixing default value for ic-admin-version

NikolaMilosa:


LittleChimera (approved):

DFINITYManu (changes requested):
I tried the URL with from-governance and yeah, it is 403.  If the code was interpolating FromGovernance into a string, and now it is interpolating from-governance, that's never gonna work.

Did you test this?

NikolaMilosa:
Of course, did you?

DFINITYManu (approved):
Now I have.  It works.  How can this fix the issue??

DFINITYManu:
Ah I see:

```

impl From<&str> for IcAdminVersion {
    fn from(value: &str) -> Self {
        match value {
            "from-governance" | "governance" | "g" | "govn" => Self::FromGovernance,
            "default" | "d" => Self::Default,
            s => Self::Strict(s.to_string()),
        }
    }
}
```

PR #895, opened 2024-09-06
refactor: `IcAdminRequirement` -> `AuthRequirement`

NikolaMilosa:
Summary of what's been done:
1. Turned `ic-admin` into lazy loaded dependency. Its fetched only if its used.
2. When developing a command we specify what auth level will be needed for it
If we are going to interact only with canisters, we need `Signer` level
If we are going to send proposals via ic-admin, we need `Neuron` level
3. Added tests for some cases

While developing i've seen noticed the sheer complexity that case `OverridenBy` adds. When that code was added we did it to ensure that if nothing is specified and we want to place a proposal for upgrading unassigned nodes we are sure to use a release automation neuron if its present on the machine. Its the only place where we've seen it be used, and in reality it was never used because that proposal is placed by Airflow. I would rather we removed this special corner case from the dre because it adds unnecessary complexity. 

**NOTE**: The above comment has been addressed in this PR and the code is removed.

I would love to try [SoftHSMv2](https://github.com/opendnssec/SoftHSMv2) to try and emulate an HSM and add tests for that detection. That could be, and probably should be a part of a different PR.


DFINITYManu:
You read my mind.     I'm looking at the parts where I could give useful input.

rs/cli/src/auth.rs
```diff
@@ -43,6 +47,162 @@ pub fn hsm_key_id_to_string(s: u8) -> String {
...
+                        Auth::Keyfile { path } if !path.exists() => Err(anyhow::anyhow!(
+                            "Path `{}` not found, which can be used to override this command. Specify your own auth args",
+                            path.display()
+                        )),
+                        Auth::Keyfile { path } => Ok((neuron.neuron_id, Auth::Keyfile { path })),
+                        // Hard error on the whole program since we don't support this
+                        _ => anyhow::bail!("Overriding neuron with auth types other than keyfile is not supported"),
+                    };
+
+                    match (maybe_user_provided_neuron, maybe_neuron_override) {
```
DFINITYManu, line 155: Do we still have a use case to preserve the override authentication branch?  What happens, hypothetically speaking, if we completely delete it?  What becomes impossible?

rs/cli/src/auth.rs
```diff
@@ -90,7 +250,7 @@ impl Neuron {
     }
 }
 
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, Eq, PartialEq)]
```
DFINITYManu, line 253: I always found it funny that Rust doesn't by default make types comparable, but when I learned that this provides the Rust compiler some opportunities for serious optimization quite literally impossible in other languages, I understood why.

rs/cli/src/auth.rs
```diff
@@ -158,6 +318,7 @@ impl Auth {
         }
     }
 
+    /// If it is called it is expected to retrieve an Auth of type Hsm or fail
```
DFINITYManu, line 321: Perfect.  I was dying to get to this point in the refinement of this function.

rs/cli/src/auth.rs
NikolaMilosa: I would love to just remove it. The use case seemed pretty good back then => If user provides nothing, assume OVERRIDE as default. Only example right now is to submit an `update_unassigned_nodes` with automation as default. Right now that part is done by Airflow and I was thinking that removing it would make the most sense because it would simplify the code, and we review proposals before submitting anyway.

rs/cli/src/auth.rs
NikolaMilosa: Was an obvious choice after the refactor

sasa-tomic (approved):
Beautiful, thanks!
(please address reviewer comments)

rs/cli/src/auth.rs
NikolaMilosa: I have decided to just remove this code. It looks way simpler and more readable. I've seen that the use case we had for this is super narrow and is a one-off.

DFINITYManu: That is grreat news!

PR #902, opened 2024-09-10
chore: adding neuron detection tests

NikolaMilosa:
This PR is intended to start adding tests for our neuron detection code with hsm. 

What is done:
1. Testing of hsm detection code for `Signer` requirement
2. Testing of hsm detection code for `Neuron` requirement
  * ATM this test fails because we can't import a dfx generated pem for a slot and use it to list neurons visible to principal and it always returns an empty vec
3. Testing of hsm detection when a specific key is provided


rs/cli/src/unit_tests/ctx_init.rs
```diff
@@ -1,4 +1,4 @@
-use std::{path::PathBuf, str::FromStr};
+use std::{path::PathBuf, process::Command, str::FromStr};
```
sasa-tomic, line 1: Do we plan to use this code for non-unit tests as well?
If so, it would make sense to rename the directory from `unit_tests` to just plain `tests` since it's shorter and does not carry less information.

sasa-tomic (approved):
TY!

rs/cli/src/unit_tests/ctx_init.rs
NikolaMilosa: I don't know. For the moment these are not even "unit tests", most are, but some do integrate with real services (Like governance canister). So what is your suggestion?

sasa-tomic: How about renaming to `tests` then?

DFINITYManu (approved):

NikolaMilosa:
Will mark as draft, still can't fully replicate hsm with softhsm

NikolaMilosa:
Will close this since its stale. We can come back to this if it makes sense in the future.

PR #903, opened 2024-09-10
feat: adding softhsm2 to docker runner

NikolaMilosa:


docker/runner.Dockerfile
```diff
@@ -87,6 +87,8 @@ USER runner
 WORKDIR /home/runner
 SHELL [ "/bin/bash", "-c" ]
 
+RUN source /opt/rye/env
```
sasa-tomic, line 90: Can this go into a separate PR?

NikolaMilosa: Okay, will wait both to be merged so I update only once, and not 2 times

LittleChimera (approved):

PR #926, opened 2024-09-13
Modify linear reduction b coefficient 70% -> 80%

pietrodimarco-dfinity:


rs/dre-canisters/trustworthy-node-metrics/src/trustworthy-node-metrics-frontend/src/components/RewardsInfo.tsx
```diff
@@ -83,7 +83,7 @@ const NodeRewardExplanation = () => {
               </ListItem>
               <ListItem sx={{ display: 'list-item' }}>
                 <Typography variant="body2" color="textSecondary" gutterBottom>
-                  Failure Rates Above 70%: Once the failure rate exceeds 70%, the rewards reduction reaches its maximum of 100%. Any failure rate beyond this threshold results in a complete loss of rewards.
+                  Failure Rates Above 80%: Once the failure rate exceeds 80%, the rewards reduction reaches its maximum of 100%. Any failure rate beyond this threshold results in a complete loss of rewards.
```
sasa-tomic, line 86: Hmmm I think we agreed with @bjoernek to saturate reductions at 80% instead of 100%, or is this something else?

PR #940, opened 2024-09-19
Prepare node rewards canister to XDR Rewards computation

pietrodimarco-dfinity:
- Added node `dc_id` `region` `node_type` `node_operator_id` infos  to nodes
- Adapted daily metrics updates to update rewards table
- Adapted daily metrics updates to update nodes with registry data
- Include new features in front end canister
- Test `node_type` computation

### DIFF
--- a/rs/dre-canisters/trustworthy-node-metrics/src/trustworthy-node-metrics/src/metrics_manager.rs
+++ b/rs/dre-canisters/trustworthy-node-metrics/src/trustworthy-node-metrics/src/metrics_manager.rs
@@ -1,5 +1,6 @@
 use std::collections::BTreeMap;
 
+use candid::Principal;
 use dfn_core::api::PrincipalId;
 use futures::FutureExt;
 use ic_base_types::NodeId;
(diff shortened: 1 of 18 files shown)


rs/dre-canisters/trustworthy-node-metrics/src/trustworthy-node-metrics/src/metrics_manager.rs
```diff
@@ -167,24 +140,121 @@ fn grouped_by_node(subnet_metrics: Vec<(PrincipalId, Vec<NodeMetricsHistoryRespo
...
+    if rewardable_nodes.is_empty() {
+        "unknown:no_rewardable_nodes_found".to_string()
+    } else {
+        loop {
+            let (k, v) = match rewardable_nodes.pop_first() {
+                Some(kv) => kv,
+                None => break "unknown:rewardable_nodes_used_up".to_string(),
+            };
+            if v != 0 {
+                break k;
```
NikolaMilosa, line 161: Don't suggest using `break` because here it is meant to actually `return`. Its misleading. `Break exp` is just a way to be able to easily assign a value from a loop, so you don't have to use `mut temp...`

NikolaMilosa (approved):

PR #942, opened 2024-09-20
feat(dre): Add support for forum post links for version elect proposals

sasa-tomic:


rs/cli/src/commands/version/revise/guest_os.rs
```diff
@@ -33,8 +33,8 @@ impl ExecutableCommand for GuestOs {
                 &ic_management_types::Artifact::GuestOs,
                 &self.version,
                 &self.release_tag,
-                self.force,
-                ctx.forum_post_link(),
+                self.ignore_missing_urls,
+                ctx.forum_post_link().expect("Forum post link is required for this proposal"),
```
NikolaMilosa, line 37: Can this go in `validate` step for this Command? It will then use clap abstractions for exiting

rs/cli/src/commands/version/revise/host_os.rs
```diff
@@ -33,8 +33,8 @@ impl ExecutableCommand for HostOs {
                 &ic_management_types::Artifact::HostOs,
                 &self.version,
                 &self.release_tag,
-                self.force,
-                ctx.forum_post_link(),
+                self.ignore_missing_urls,
+                ctx.forum_post_link().expect("Forum post link is required for this proposal"),
```
NikolaMilosa, line 37: Same as previous

NikolaMilosa (approved):
Amazing, thanks! Just a couple of nits

rs/cli/src/commands/version/revise/guest_os.rs
sasa-tomic: I wanted to have that, yes. But validate does not have access to the command, only to ctx, and the refactor to get the into the validate isn't quite straightforward. But I do feel we should make that change, so I agree with you.

sasa-tomic: fixed, PTAL

PR #943, opened 2024-09-20
refactor(dre): renaming to `no_sync` to `offline` for clarity

NikolaMilosa:
"No sync" is the term we use only for syncing with registry canister. This flag is intended to have a broader use which covers the case of when public dashboard is offline as well.

### DIFF
--- a/rs/cli/src/commands/mod.rs
+++ b/rs/cli/src/commands/mod.rs
@@ -194,7 +194,7 @@ The argument is mandatory for testnets, and is optional for mainnet and staging"
     #[clap(long, env = "VERBOSE", global = true)]
     pub verbose: bool,
 
-    /// Run the tool offline when possible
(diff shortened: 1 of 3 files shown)


rs/cli/src/commands/mod.rs
```diff
@@ -194,11 +194,11 @@ The argument is mandatory for testnets, and is optional for mainnet and staging"
     #[clap(long, env = "VERBOSE", global = true)]
     pub verbose: bool,
 
-    /// Don't sync with the registry
+    /// Run the tool offline when possible
```
sasa-tomic, line 197:
```suggestion
    /// Run the tool offline when possible, i.e., do not sync registry and public dashboard data before the run
```

sasa-tomic (approved):

PR #955, opened 2024-09-23
chore: removing unused parts of code

NikolaMilosa:
This PR focuses on removing obsolete and stale code:
1. `np-notifications` - an old architecture for sending notifications to node providers
2. `rollout-controller` - a failed attempt to replace what we currently have with something easier to use
3. `qualifier` - obsolete tool which is replaced by a qualification test in ic repo


sasa-tomic (approved):
Whoaaa! Beautiful!

PR #961, opened 2024-09-25
feat(dre): Improve dre registry a bit more

sasa-tomic:
- Add a bit more filtering operator beside '=', such as '!=', '>', '<', startswith, endswith, contains, etc.
- Remove the now unnecessary --incorrect-rewards command line option, replaced by: `--filter rewards_correct=false`
- Add node details in the subnet list, for convenience (albeit this adds some verbosity, so if it's too verbose we can remove it in the future)
- Reorder the output fields a bit, putting the list of subnets first

### DIFF
--- a/rs/cli/src/commands/registry.rs
+++ b/rs/cli/src/commands/registry.rs
@@ -19,6 +19,7 @@ use ic_types::PrincipalId;
 use indexmap::IndexMap;
 use itertools::Itertools;
 use log::{info, warn};
+use regex::Regex;
 use serde::Serialize;
 use serde_json::Value;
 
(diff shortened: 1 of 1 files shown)


rs/cli/src/commands/registry.rs
```diff
@@ -317,6 +338,10 @@ async fn get_nodes(
     Ok(nodes)
 }
 
+fn get_unassigned_nodes(local_registry: &Arc<dyn LazyRegistry>) -> anyhow::Result<Option<UnassignedNodesConfigRecord>> {
+    local_registry.get_unassigned_nodes()
+}
+
```
NikolaMilosa, line 344: At this moment, we should remove it, its an empty function. I forgot to remove it in my refactoring.

NikolaMilosa (approved):

NikolaMilosa:
Maybe add a couple of examples in description of the PR how to invoke it so we know what the idea was because it changed

rs/cli/src/commands/registry.rs
sasa-tomic: done

PR #964, opened 2024-09-26
chore: release `rc--2024-09-26_01-31`

NikolaMilosa:


NikolaMilosa:
In draft because we are awaiting the qualification of `ubuntu24` feature build and the `no-canister-snapshots` backup

sasa-tomic (approved):

PR #966, opened 2024-09-27
feat: cordon features

NikolaMilosa:
This PR adds the ability to cordon features by adding entries to the new `cordoned_features.yaml` file. Currently it will exclude from the pool of available nodes all the nodes with the cordoned feature. It will not attempt to remove all healthy nodes (in the subnet) that have said feature, it will just prevent entering of the new nodes with those features.

When removing nodes from the subnet and when they are added back to the available pool, if the removed node contains cordoned features it will not be added to the available pool. It means that some of the nodes containing cordoned features _could_ be removed if the business rules are met.

### DIFF
--- a/rs/cli/Cargo.toml
+++ b/rs/cli/Cargo.toml
@@ -77,6 +77,7 @@ comfy-table = { workspace = true }
 human_bytes = { workspace = true }
 mockall.workspace = true
 clio = { workspace = true }
+serde_yaml.workspace = true
 
 [dev-dependencies]
 actix-rt = { workspace = true }
--- a/rs/cli/src/cordoned_feature_fetcher.rs
+++ b/rs/cli/src/cordoned_feature_fetcher.rs
@@ -1,0 +1,126 @@
+use std::{path::PathBuf, time::Duration};
+
+use decentralization::network::NodeFeaturePair;
+use futures::future::BoxFuture;
+use ic_management_types::NodeFeature;
+use itertools::Itertools;
+use log::warn;
+use mockall::automock;
+use reqwest::{Client, ClientBuilder};
+use strum::VariantNames;
+
+#[automock]
+pub trait CordonedFeatureFetcher: Sync + Send {
+    fn fetch(&self) -> BoxFuture<'_, anyhow::Result<Vec<NodeFeaturePair>>>;
+}
+
+pub struct CordonedFeatureFetcherImpl {
+    client: Client,
+    fallback_file: Option<PathBuf>,
+    offline: bool,
+}
+
+const CORDONED_FEATURES_FILE_URL: &str = "https://raw.githubusercontent.com/dfinity/dre/refs/heads/main/cordoned_features.yaml";
+
+impl CordonedFeatureFetcherImpl {
+    pub fn new(offline: bool, fallback_file: Option<PathBuf>) -> anyhow::Result<Self> {
+        let client = ClientBuilder::new().timeout(Duration::from_secs(10)).build()?;
+
+        Ok(Self {
+            client,
+            fallback_file,
+            offline,
+        })
+    }
+
+    async fn fetch_from_git(&self) -> anyhow::Result<Vec<NodeFeaturePair>> {
+        let bytes = self
+            .client
+            .get(CORDONED_FEATURES_FILE_URL)
+            .send()
+            .await?
+            .error_for_status()?
+            .bytes()
+            .await?;
+
+        self.parse(&bytes)
+    }
+
+    fn fetch_from_file(&self, path: &PathBuf) -> anyhow::Result<Vec<NodeFeaturePair>> {
+        let contents = std::fs::read(path)?;
+
+        self.parse(&contents)
+    }
+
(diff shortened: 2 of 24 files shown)


rs/cli/Cargo.toml
```diff
@@ -77,6 +77,7 @@ comfy-table = { workspace = true }
 human_bytes = { workspace = true }
 mockall.workspace = true
 clio = { workspace = true }
+serde_yaml.workspace = true
```
sasa-tomic, line 80: We should be able to easily switch to https://github.com/acatton/serde-yaml-ng in one of the future PRs.

rs/cli/src/cordoned_feature_fetcher.rs
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    fn fetch_from_file(&self, path: &PathBuf) -> anyhow::Result<Vec<NodeFeaturePair>> {
+        let contents = std::fs::read(path)?;
+
+        self.parse(&contents)
+    }
+
+    // Write tests for this
+    fn parse(&self, contents: &[u8]) -> anyhow::Result<Vec<NodeFeaturePair>> {
```
sasa-tomic, line 56: This function doesn't seem to need `&self`.

rs/cli/src/cordoned_feature_fetcher.rs
```diff
@@ -0,0 +1,174 @@
...
+            .send()
+            .await?
+            .error_for_status()?
+            .bytes()
+            .await?;
+
+        self.parse(&bytes)
+    }
+
+    fn fetch_from_file(&self, path: &PathBuf) -> anyhow::Result<Vec<NodeFeaturePair>> {
```
sasa-tomic, line 49: Also doesn't need self if `parse` doesn't.

rs/cli/src/cordoned_feature_fetcher.rs
```diff
@@ -0,0 +1,174 @@
...
+        let client = ClientBuilder::new().timeout(Duration::from_secs(10)).build()?;
+
+        Ok(Self {
+            client,
+            fallback_file,
+            offline,
+        })
+    }
+
+    async fn fetch_from_git(&self) -> anyhow::Result<Vec<NodeFeaturePair>> {
```
sasa-tomic, line 36: Same as below - doesn't need self if `parse` doesn't.

rs/cli/src/cordoned_feature_fetcher.rs
```diff
@@ -0,0 +1,174 @@
...
+        }
+
+        Ok(valid_features)
+    }
+}
+
+impl CordonedFeatureFetcher for CordonedFeatureFetcherImpl {
+    fn fetch(&self) -> BoxFuture<'_, anyhow::Result<Vec<NodeFeaturePair>>> {
+        Box::pin(async {
+            match (self.offline, self.fallback_file.as_ref()) {
```
sasa-tomic, line 105: Nice, a `match` here makes the code very readable!

rs/cli/src/unit_tests/replace.rs
```diff
@@ -0,0 +1,345 @@
...
+        subnet.id,
+        subnet.nodes.iter().map(|node| pretty_print_node(node, 2)).join("\n")
+    )
+}
+
+#[test]
+fn should_skip_cordoned_nodes() {
+    // World setup
+    let available_nodes = vec![
+        node(1, true, &[(NodeFeature::Country, "Country 1"), (NodeFeature::City, "City 1")]),
```
sasa-tomic, line 129: We're getting rid of the "NodeFeature::City" since it clashes with the adopted target topology. Can you use DC instead?

sasa-tomic (approved):

sasa-tomic (approved):

rs/cli/src/cordoned_feature_fetcher.rs
NikolaMilosa: True! What I like about all these cases is that I can do `self.parse(...)` rather than `Self::parse()`. It makes the code more readable to me so that is why I prefer it. If you would rather me remove it all do it 😄

rs/cli/src/cordoned_feature_fetcher.rs
NikolaMilosa: It does need it because of the client config.

rs/cli/src/cordoned_feature_fetcher.rs
NikolaMilosa: I made a logical error here, will refactor the code a bit, here it should use the path from `self`

rs/cli/src/cordoned_feature_fetcher.rs
sasa-tomic: It's one less argument that needs to be passed to the function, but it's not a big deal either way. We are not performance sensitive in this call. So it's fine if we leave it as is.

sasa-tomic (approved):
Awesome, thank you!

PR #972, opened 2024-09-27
Add EU and remove Continent from calculation of Nakamoto

sasa-tomic:
Add the EU and remove the Continent from the calculation of Nakamoto, as [requested in the forum](https://forum.dfinity.org/t/subnet-management-uzr34-ii/33981/16?u=sat) and reviewed and approved internally (BjörnA, Sven, Sam).

Also rename the `City` into the `Area` field in the `Datacenter` struct, skipping the `Continent` feature when iterating over the `NodeFeature` enum members, and adjusting the mapping of EU countries in the `From<&ic_management_types::Node> for Node` implementation for easier readability. For EU countries, we replace the City/Area with the Country, to keep EU in line with the US decentralization info, where we have the "State" in the City/Area.


LittleChimera (approved):

aodl:
I've not had time to do much more than glance at the code, but it look and sounds good. Thanks Sat!

PR #977, opened 2024-10-02
feat: lazy auth

NikolaMilosa:
This PR refactors the way we do proposals. Now the following rule applies:
> If the detection of neuron fails a fake neuron will be put in its place. That neuron doesn't exist on the mainnet and cannot submit proposals. 

This is done in order to allow the team to play around with different variations of commands without having to always say `--dry-run`. 

When the hsm is plugged in, or when the private key pem is specified the code will use that neuron.


rs/cli/src/auth.rs
```diff
@@ -87,8 +87,29 @@ impl Neuron {
         Self::from_opts_and_req_inner(auth_opts, requirement, network, neuron_id).await
     }
 
-    pub async fn dry_run_fake_neuron(network: &Network) -> anyhow::Result<Self> {
-        Self::from_opts_and_req(AuthOpts::none(), AuthRequirement::Anonymous, network, None).await
+    fn ensure_fake_pem(name: &str) -> anyhow::Result<PathBuf> {
```
sasa-tomic, line 90:
```suggestion
    fn new_with_fake_pem(name: &str) -> anyhow::Result<PathBuf> {
```

rs/cli/src/auth.rs
```diff
@@ -87,8 +87,29 @@ impl Neuron {
...
+        if !path.exists() {
+            std::fs::write(&path, "Some private key")?;
+        }
+        Ok(path)
+    }
+
+    pub fn dry_run_fake_neuron() -> anyhow::Result<Self> {
+        Ok(Self {
+            auth: Auth::Keyfile {
+                path: Self::ensure_fake_pem("test_neuron_1")?,
```
sasa-tomic, line 108:
```suggestion
                path: Self::new_with_fake_pem("test_neuron_1")?,
```

rs/cli/src/auth.rs
nikolamilosasotex: But it just ensures that a file on that path exists. It doesn't return `Self`?

NikolaMilosa: But it just ensures that a path exists. It doesn't return `Self`?

sasa-tomic: ah right, my bad, sorry

rs/cli/src/auth.rs
sasa-tomic: yep, sorry

sasa-tomic (approved):

PR #985, opened 2024-10-04
chore: new runner 

NikolaMilosa:
Updating runner version and zig in that version

### DIFF
--- a/docker/runner.Dockerfile
+++ b/docker/runner.Dockerfile
@@ -12,7 +12,7 @@ RUN apt-get update && \
     libssl-dev zlib1g-dev libbz2-dev libreadline-dev libsqlite3-dev libffi-dev liblzma-dev libncurses5-dev libncursesw5-dev \
     xz-utils tk-dev libffi-dev liblzma-dev python-openssl protobuf-compiler libdbus-1-dev softhsm2 libsofthsm2 opensc -y
 
-RUN curl -L https://ziglang.org/builds/zig-linux-x86_64-0.14.0-dev.321+888708ec8.tar.xz | tar -xJ && \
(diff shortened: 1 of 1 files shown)


docker/runner.Dockerfile
```diff
@@ -12,7 +12,7 @@ RUN apt-get update && \
     libssl-dev zlib1g-dev libbz2-dev libreadline-dev libsqlite3-dev libffi-dev liblzma-dev libncurses5-dev libncursesw5-dev \
     xz-utils tk-dev libffi-dev liblzma-dev python-openssl protobuf-compiler libdbus-1-dev softhsm2 libsofthsm2 opensc -y
 
-RUN curl -L https://ziglang.org/builds/zig-linux-x86_64-0.14.0-dev.321+888708ec8.tar.xz | tar -xJ && \
+RUN curl -L https://ziglang.org/builds/zig-linux-x86_64-0.14.0-dev.1713+f99097e68.tar.xz | tar -xJ && \
```
sasa-tomic, line 15: Can we use the (more) stable 0.13 instead?
`https://ziglang.org/download/0.13.0/zig-linux-x86_64-0.13.0.tar.xz`

sasa-tomic, line 15: Can we use the (more) stable 0.13 instead?
`https://ziglang.org/download/0.13.0/zig-linux-x86_64-0.13.0.tar.xz`

NikolaMilosa: I've put 0.14

sasa-tomic: I know but that's a dev build from some random revision, which obviously gets GCed after some time. Can we use 0.13 instead that is still available upstream, so is more likely not to disappear that quickly?

NikolaMilosa: Oh its not stable. Sorry.... Changing... 😟

sasa-tomic (approved):

PR #998, opened 2024-10-08
fix: using different api for pagination

NikolaMilosa:
At the moment we have an issue with the discourse client fetching only 20 posts (which is the default). This results in crashing and recreating the same message
> We’re preparing a new IC release. The changelog will be announced soon.

Sadly the [DiscourseClient](https://github.com/pydiscourse/pydiscourse/blob/master/src/pydiscourse/client.py) we use doesn't allow for using any suggestions from [this blog post about pagination](https://meta.discourse.org/t/fetch-all-posts-from-a-topic-using-the-api/260886) so I had to tweak the code and use standard `requests` for this.

### DIFF
--- a/release-controller/forum.py
+++ b/release-controller/forum.py
@@ -27,6 +27,28 @@ Here is a summary of the changes since the last release:
 """
 
 
+class RequestsPostFetcher:
+    """Posts fetcher using requests."""
+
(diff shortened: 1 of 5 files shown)


NikolaMilosa:
Not needed because of #1000

PR #999, opened 2024-10-09
Explain that the reproduction process is only important for GuestOS.

DFINITYManu:


LittleChimera (changes requested):
we probably want to hold this off until we have proper changes in place

release-controller/release_index_loader.py
```diff
@@ -23,7 +23,7 @@ def _verify_release_instructions(version):
 sudo apt-get install -y curl && curl --proto '=https' --tlsv1.2 -sSLO https://raw.githubusercontent.com/dfinity/ic/{version}/ci/tools/repro-check.sh && chmod +x repro-check.sh && ./repro-check.sh -c {version}
 ```
 
-The two SHA256 sums printed above from a) the downloaded CDN image and b) the locally built image, must be identical, and must match the SHA256 from the payload of the NNS proposal.
+The two SHA256 sums printed above from a) the downloaded CDN image and b) the locally built image, must be identical, and must match the SHA256 from the payload of the NNS proposal.  The verification process will also attempt to reproduce HostOS and SetupOS — these results are for advisory purposes only, as this proposal only elects a GuestOS.
```
sasa-tomic, line 26: We have been discussing submitting both the HostOS and GuestOS proposal together on a weekly basis, so let's wait with merging this PR until we decide what to do on that matter.

sasa-tomic (changes requested):

sasa-tomic:
Implemented similar changes in https://github.com/dfinity/dre/pull/1028

PR #1000, opened 2024-10-09
implement pagination to fetch created posts

LittleChimera:


release-controller/forum.py
```diff
@@ -74,13 +76,19 @@ def __init__(
...
-        topic_posts = self.client.topic_posts(topic_id=self.topic_id)
-        if not topic_posts:
-            raise RuntimeError("failed to list topic posts")
-
-        return [
-            p for p in topic_posts.get("post_stream", {}).get("posts", {}) if p["yours"]
-        ]
+        results = []
+        for p in range((self.posts_count - 1) // 20 + 1):
+            topic_posts = self.client._get(f"/t/{self.topic_id}.json", page=p + 1)
```
NikolaMilosa, line 81: If you can use `client._get` why don't you just add `print=true`? It will fetch 1000 posts which should in 99.99% be enough. This requires more fetches and it might get rate limited 😟

NikolaMilosa: Discussed privately

NikolaMilosa (approved):

PR #1006, opened 2024-10-14
feat(dre): offline mode

NikolaMilosa:
This PR adds a centralized local store for various artifacts required by `dre` to run properly. The goal is to provide local cache for all things we download off the internet (computer) which would allow the tool to run to some extent without any connection. 

On top of that it further enhances the testability of the tool giving us the ability to easily substitute files we need in tests.


rs/cli/src/store.rs
```diff
@@ -0,0 +1,273 @@
...
+    }
+
+    fn cordoned_features_file(&self) -> anyhow::Result<PathBuf> {
+        let file = self.path().join("cordoned_features.yaml");
+
+        if !file.exists() {
+            info!("Cordoned features file was missing. Creating on path `{}`...", file.display());
+            if let Err(e) = std::fs::write(&file, "") {
+                warn!("Failed to create cordoned features file: {:?}", e);
+                warn!("This is not critical now. If github is offline then it could be");
```
sasa-tomic, line 262:
```suggestion
                warn!("This is not a critical error, but would be if GitHub is unreachable.");
```

rs/cli/src/unit_tests/ctx_init.rs
```diff
@@ -4,17 +4,22 @@ use crate::{
...
 use itertools::Itertools;
 
-use crate::{commands::IcAdminVersion, ctx::DreContext, ic_admin::FALLBACK_IC_ADMIN_VERSION};
+use crate::{commands::IcAdminVersion, ctx::DreContext};
 
 fn status_file_path() -> PathBuf {
-    dirs::home_dir().unwrap().join("bin").join("ic-admin.revisions").join("ic-admin.status")
+    dirs::cache_dir()
+        .unwrap()
+        .join("dre-store")
```
sasa-tomic, line 20: Can we have this directory in a single place only? E.g. there could be a class method that returns the local store path, and would be called both from the `Store` and from here.

sasa-tomic (approved):
Looks great, and looking forward to the tests for the new functionality! :D

rs/cli/src/unit_tests/ctx_init.rs
NikolaMilosa: Tests are not finished, so this will change soon when I start adding tests for the agreed implementation

rs/cli/src/store.rs
NikolaMilosa: This whole log doesn't make sense because that code wouldn't be reached if github is down. I will write a clearer message.

rs/cli/src/auth.rs
```diff
@@ -82,7 +83,12 @@ impl Neuron {
             (neuron_id, auth_opts)
         };
 
-        Self::from_opts_and_req_inner(auth_opts, requirement, network, neuron_id).await
+        Self::from_opts_and_req_inner(auth_opts, requirement, network, neuron_id, offline).await
+    }
+
+    #[cfg(test)]
+    pub fn ensure_fake_pem_outter(name: &str) -> anyhow::Result<PathBuf> {
```
sasa-tomic, line 90:
```suggestion
    pub fn ensure_fake_test_pem(name: &str) -> anyhow::Result<PathBuf> {
```

rs/cli/src/store.rs
```diff
@@ -0,0 +1,298 @@
...
+                dir.display()
+            );
+            std::fs::create_dir_all(&dir)?
+        }
+
+        Ok(dir)
+    }
+
+    #[cfg(test)]
+    pub fn guest_labels_cache_path_outter(&self, network: &Network) -> anyhow::Result<PathBuf> {
```
sasa-tomic, line 92:
```suggestion
    pub fn guest_labels_cache_path_outer(&self, network: &Network) -> anyhow::Result<PathBuf> {
```

outter ==> outer

although, having a "test" in the name is clearer for me, since the purpose is clearer.

rs/cli/src/store.rs
```diff
@@ -0,0 +1,298 @@
...
+        if !path.exists() {
+            info!("ic-admin.revisions dir was missing. Creating on path `{}`...", path.display());
+            std::fs::create_dir_all(&path)?;
+        }
+
+        Ok(path)
+    }
+
+    #[cfg(test)]
+    pub fn ic_admin_status_file_outter(&self) -> anyhow::Result<PathBuf> {
```
sasa-tomic, line 139: same, outter ==> outer (or test)

sasa-tomic (approved):

PR #1029, opened 2024-10-22
feat: testing to make `get_certified_changes_since` and `get_changes_since` the same

NikolaMilosa:


NikolaMilosa:
This was just testing the differences between calls

PR #1068, opened 2024-11-11
feat: ensure that there are some publishable changes in release notes

NikolaMilosa:
This PR modifies the step before creating a PR on our repository for reviewing changes. Previously the only criteria we checked was if all the teams reviewed the notes (by crossing their name out). With this PR we will check how many changes are publishable. 

Change is publishable if its not: 
* Manually excluded by developers
* Automatically excluded by release controller

### DIFF
--- a/release-controller/publish_notes.py
+++ b/release-controller/publish_notes.py
@@ -8,6 +8,8 @@ from github import Github
 from github.Repository import Repository
 import markdown
 from itertools import groupby
+from google_docs import ReleaseNotesClient
+import pathlib
 
 REPLICA_RELEASES_DIR = "replica-releases"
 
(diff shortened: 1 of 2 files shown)


release-controller/publish_notes.py
```diff
@@ -121,16 +123,54 @@ def publish_if_ready(self, google_doc_markdownified, version: str):
             return
 
         changelog = changelog[release_notes_start:]
+        if check_number_of_changes(changelog) == 0:
+            logging.warning("release notes for version %s contain no commits that would be published.")
```
sasa-tomic, line 127: This behavior sounds like an "error" rather than a "warning"?

DFINITYManu (approved):

PR #1074, opened 2024-11-11
Add NPR lib to NPR App

pietrodimarco-dfinity:
This adds the NPR library to the NPR canister removing the rewards computation from the canister itself.
Other changes included in the PR:

- Add CandidType to result structs of the lib.
- Add log level for `RewardsLog` to not always print entire execution
- Remove unused in canister logger


sasa-tomic (approved):
LGTM

PR #1080, opened 2024-11-13
feat(decentralization): from nodes with same dec. prefer those from less used operators

sasa-tomic:
From all nodes that result in the same decentralization (Nakamoto coefficients) prefer those nodes that come from less used node operators. This should balance out the node operator utilization, and overall result in better decentralization across all subnets.


pietrodimarco-dfinity (approved):
THX!

PR #1090, opened 2024-11-15
Add machines limit for not penalise NP

pietrodimarco-dfinity:
- Fix bug in cache to create parent if path to `/Library/Caches/ic-registry-cache` doesn't exist
- Add `FULL_REWARDS_MACHINES_LIMIT` per NP


rs/dre-canisters/node-provider-rewards-lib/src/v1_rewards.rs
```diff
@@ -231,43 +232,54 @@ fn node_provider_rewards(
...
+            region_nodetype_rewards.get(&region_key).expect("Type3 rewards already filled")
+        } else {
+            region_nodetype_rewards
+                .get(&(region.clone(), node_type.clone()))
+                .expect("Rewards already filled")
+        };
+        let rewards_xdr_no_penalty = Operation::Multiply(*xdr_permyriad, Decimal::from(*node_count));
+        rewards_xdr_no_penalty_total.push(rewards_xdr_no_penalty.clone());
+
+        // Node Providers with less than 4 machines are rewarded fully, independently on their performance
```
sasa-tomic, line 256:
```suggestion
        // Node Providers with less than 4 machines are rewarded fully, independently of their performance
```

rs/ic-management-backend/src/public_dashboard.rs
```diff
@@ -2,6 +2,7 @@ use crate::registry::local_cache_path;
 use ic_management_types::Network;
 use serde::de::DeserializeOwned;
 use std::path::PathBuf;
+use tokio::fs;
```
sasa-tomic, line 5: https://docs.rs/fs-err/latest/fs_err/ is a good idea instead of `fs`, since `fs` does not have path information in case of errors.

sasa-tomic (approved):

rs/ic-management-backend/src/public_dashboard.rs
pietrodimarco-dfinity: Noice :)

PR #1093, opened 2024-11-15
Allow DRE vote and other operations to repair the PKCS#11 context that might have gone bad.

DFINITYManu:
When suspending a Linux system, or when accidentally touching a loose NitroKey, the PKCS#11 context used to talk to it gets broken and does not know how to recover.  When we attempt to sign messages, we now detect this situation, drop the prior context, and create a new one.  If this succeeds, all good.  If this fails, we report the original error to the user, and notify them as well that we could not regenerate a valid PKCS#11 context.

In practice, this means that `dre vote` will print an error and send a notification if it can't continue, but if a PKCS#11 error was successfully fixed behind the scenes, it will stay in the loop and continue working fine.


pietrodimarco-dfinity (approved):

sasa-tomic (approved):
nice!

PR #1098, opened 2024-11-21
feat(dre): Discourse client for automatic forum post management around proposals.

NikolaMilosa:
This pull request adds an initial discourse client. The discourse client takes care of low level calls and high level coordination should be implemented into a Trait which exposes a clear api.

At the moment only the `subnet replace` command is changed to use the discourse api. If this was to be merged, further code changes for relevant proposal types would be added in a separate PR.

Still thinking about how to reliably test it.

### DIFF
--- a/rs/cli/src/commands/mod.rs
+++ b/rs/cli/src/commands/mod.rs
@@ -131,7 +131,7 @@ pub struct DiscourseOpts {
     #[clap(long, env = "DISCOURSE_API_KEY")]
     pub(crate) discourse_api_key: Option<String>,
 
-    /// Api key used to interact with the forum
+    /// Api user that will interact with the forum
     #[clap(long, env = "DISCOURSE_API_USER")]
     pub(crate) discourse_api_user: Option<String>,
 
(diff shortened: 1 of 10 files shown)


rs/cli/src/commands/mod.rs
```diff
@@ -125,12 +125,30 @@ impl TryFrom<String> for AuthOpts {
     }
 }
 
+#[derive(ClapArgs, Debug, Clone)]
+pub struct DiscourseOpts {
+    /// Api key used to interact with the forum
+    #[clap(long, env = "DISCOURSE_API_KEY")]
+    pub(crate) discourse_api_key: Option<String>,
+
+    /// Api key used to interact with the forum
```
sasa-tomic, line 134:
```suggestion
    /// Api user that will interact with the forum
```

sasa-tomic (approved):
LGTM! Thank you!

rs/cli/src/commands/mod.rs
NikolaMilosa: Nice catch!

PR #1106, opened 2024-11-22
Update target_determinator to v0.28.0

pietrodimarco-dfinity:


sasa-tomic:
Seems to be the same as #1104

PR #1122, opened 2024-11-28
chore(IDX): remove custom token

cgundy:
It does not appear that you actually needed a custom token here, so I am removing it.


sasa-tomic (approved):
Let's give it a shot

PR #1135, opened 2024-12-04
refactor(): split into 2 jobs

cgundy:
This splits the workflow into two separate jobs. The benefit is that permissions can be defined separately and the jobs can be parallelized.


sasa-tomic:
Closing in favor of https://github.com/dfinity/dre/pull/1136

PR #1136, opened 2024-12-04
fix(ci): use GitHub App token to trigger CI actions

sasa-tomic:
An alternative implementation similar to https://github.com/dfinity/dre/pull/1135 would also work, but that would approximately double the load on the CI.
The change from this PR doesn't increase the load, although admitedly it's not as clean as #1135

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -36,7 +36,7 @@ jobs:
           fetch-depth: 2
 
       # The GitHub App token is necessary for pushing changed files back to the repository
-      # If secrets.GITHUB_TOKEN is used, the push will not trigger any actions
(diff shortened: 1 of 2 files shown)


NikolaMilosa (approved):

.github/workflows/main.yaml
```diff
@@ -34,6 +34,17 @@ jobs:
       - uses: actions/checkout@v4
         with:
           fetch-depth: 2
+
+      # The GitHub App token is necessary for pushing changed files back to the repository
+      # If secrets.GITHUB_TOKEN is used, the push will not trigger any actions
```
sasa-tomic, line 39:
```suggestion
      # If regular secrets.GITHUB_TOKEN is used instead, the push will not trigger any actions
```

PR #1146, opened 2024-12-09
chore: reordering nav for github pages

NikolaMilosa:
This PR reorders navigation on github pages and includes a new "Chatbot" page and adds it to the top of our docs. It will render an IFrame to the support bot that we use for user's to submit issues.


sasa-tomic (approved):
LGTM

PR #1148, opened 2024-12-11
refactor: Refactoring discourse to create posts to a single topic per subnet

NikolaMilosa:
Previously we used to create a new topic for each proposal. With this PR we will create posts in already created topics per subnet.


rs/cli/src/discourse_client.rs
```diff
@@ -149,32 +179,54 @@ impl DiscourseClientImp {
...
+            .interact()?;
+        Ok(Some(DiscourseResponse {
+            url: forum_post_link,
+            update_id: None,
+        }))
+    }
 }
 
 const GOVERNANCE_TOPIC: &str = "Governance";
 const SUBNET_MANAGEMENT_TAG: &str = "Subnet-management";
```
sasa-tomic, line 199: We need one more tag here, it was mentioned in one of the threads somewhere.

rs/cli/src/discourse_client.rs
```diff
@@ -149,32 +179,54 @@ impl DiscourseClientImp {
...
             if self.offline || self.skip_forum_post_creation {
                 return Ok(None);
             }
-            let topic = self.create_topic(post).await?;
-            Ok(Some(topic))
-        };
-        Box::pin(async move { try_call.await.or_else(|e| self.request_from_user(e, post_clone)) })
+
+            let subnet_topic_map = get_subnet_topics_map();
+            let topic_info = subnet_topic_map.get(&subnet_id).ok_or(anyhow::anyhow!(
```
sasa-tomic, line 220: Would be good to let the user override this default with a cli command, and we actually already have a cli option we could use here.

rs/cli/src/discourse_client.rs
NikolaMilosa: I believe we cannot tag specific posts in a topic. The topics that exist are already properly tagged.

rs/cli/src/discourse_client.rs
NikolaMilosa: Do we want to allow the override? When would we want a replace proposal to be sent to a different topic?

rs/cli/src/discourse_client.rs
sasa-tomic: These tags seem to be still used for authorized subnets proposals.
I've found the relevant doc in the meantime: https://docs.google.com/document/d/1Aar4w0JnUZTfXIaBp4PHvK7BP4G4s5IE6JxkuSfbycs/edit?pli=1&tab=t.0#heading=h.yd166uqdjfui

We should also add the subcategory "NNS proposal discussions" to these topics.

rs/cli/src/discourse_client.rs
sasa-tomic: I can envision some cases, like when some community members request a node replacement in some forum thread, and would like to link to the specific forum post. However, in that case we probably don't want to modify the forum post upon proposal submission - especially if it's one of NP/community member posts.

rs/cli/src/discourse_client.rs
NikolaMilosa: Should that be a part of this PR? I can add it but seems like it should redone in another PR?

sasa-tomic: Sure, another PR sounds good.

sasa-tomic (approved):
TY!

PR #1159, opened 2024-12-19
feat: Adding possibility to override neurons per command

NikolaMilosa:
There are some proposals that we would like to _automatically_ submit with correct neuron that is not the default neuron (Like subnet version upgrade proposals). This PR adds the possibility to add an override on a command level that will be used only if no other auth opts are specified.

This PR also includes removing of some boiler plate code as well and turns it into a macro


sasa-tomic (approved):
LGTM

PR #1178, opened 2025-01-06
chore(dre): New defaults for state size when updating authorized subnets

NikolaMilosa:
After recent development the limits adjusted. During discussion it was noted that state size criteria technically could be removed all together but we opted to keep it until further CMC development takes place.

Full discussion in [this thread](https://dfinity.slack.com/archives/C01DB8MQ5M1/p1736151170753039).

### DIFF
--- a/rs/cli/src/commands/update_authorized_subnets.rs
+++ b/rs/cli/src/commands/update_authorized_subnets.rs
@@ -17,7 +17,7 @@ use crate::{
 use super::{AuthRequirement, ExecutableCommand};
 
 const DEFAULT_CANISTER_LIMIT: u64 = 60_000;
-const DEFAULT_STATE_SIZE_BYTES_LIMIT: u64 = 429_496_729_600; // 400GB
+const DEFAULT_STATE_SIZE_BYTES_LIMIT: u64 = 400 * 1024 * 1024 * 1024; // 400GB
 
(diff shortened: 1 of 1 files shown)


rs/cli/src/commands/update_authorized_subnets.rs
```diff
@@ -17,7 +17,7 @@ use crate::{
 use super::{AuthRequirement, ExecutableCommand};
 
 const DEFAULT_CANISTER_LIMIT: u64 = 60_000;
-const DEFAULT_STATE_SIZE_BYTES_LIMIT: u64 = 322_122_547_200; // 300GB
+const DEFAULT_STATE_SIZE_BYTES_LIMIT: u64 = 429_496_729_600; // 400GB
```
sasa-tomic, line 20:
```suggestion
const DEFAULT_STATE_SIZE_BYTES_LIMIT: u64 = 400 * 1024 * 1024 * 1024
```

sasa-tomic (approved):

rs/cli/src/commands/update_authorized_subnets.rs
sasa-tomic: also:
400GB == 400_000_000_000
400GiB == 400 * 1024^3

PR #1179, opened 2025-01-06
fix: boundary nodes update script migration to github

NikolaMilosa:
It turns out this script is still needed so here is a fix for its migration to github from gitlab.


sasa-tomic (approved):
LGTM

PR #1194, opened 2025-01-13
Support security hotfixes in the release controller.

DFINITYManu:
Fixes https://dfinity.atlassian.net/browse/DRE-219 .

Effectively, we skip changelog generation from Git for security hotfixes, replacing them with the text:

> In accordance with the Security Patch Policy and Procedure that was adopted in proposal
> 48792, the source code that was used to build this release will be exposed at the latest
> 10 days after the fix is rolled out to all subnets.
>
> The community will be able to retroactively verify the binaries that were rolled out.

**No review checklist is created in the Google doc.**  This almost certainly means that the release will proceed right away, with no one to be notified.  **If you don't want this, add a review indicating the team to be notified, the Slack ID of the team, and the Google docs @ handle.**

The rest of the process remains identical.


LittleChimera (approved):

DFINITYManu:
Thanks for the approval @LittleChimera.  I will hold off on merging this to create a dry-run mode that allows me to test the reconciler changes.

PR #1200, opened 2025-01-14
fix: revert git cliff and bump aspect lib 

NikolaMilosa:
Since current setup isn't sufficient for what we need it to do this PR reverts `git-cliff` for now. 

Since the cache was wiped it was seen that the version `2.6.0` for aspect lib was yanked so thins PR also bumps the version of it to latest (`2.11.0`)


NikolaMilosa:
Not needed as #1201 will fix the issue introduced here

PR #1209, opened 2025-01-15
chore: using `NNS proposal discussion` category for creating new posts

NikolaMilosa:
Discourse treats subcategories as categories that have a `parent` but returns a weird structure that is recursive. After this PR the proposals that are automatically created by the dre tool will have a category looking like the following:
image


sasa-tomic (approved):
Fantastic, thank you!

PR #1213, opened 2025-01-16
Enhancements to release controller

DFINITYManu:
1. Full typification of release controller.
2. Functional change: check on start for existing election proposals, and save them in local database.
3. Functional change: only do the work of preparing and submitting a proposal when a proposal has not successfully been submitted in the past.  This is bypassable on the command line using `--skip-preloading-state` and a clean reconciler state store.
4. Enhanced dry-run mode.
5. Now with progress bar whenever checksumming the GuestOS distribution packages.
6. No longer write the enormous GuestOS distribution packages to disk in order to checksum.  Much faster checksumming as a result.
7. Only fetch git notes at most once per 30 seconds.  Much faster preparation of release notes as a result.
8. Option to loop only one time (`--loop-every=0`) and to adjust how often the reconciliation process shall loop.


BUILD.bazel
```diff
@@ -34,4 +34,5 @@ py_console_script_binary(
         requirement("mkdocs-git-revision-date-localized-plugin"),
         requirement("mkdocs-git-committers-plugin-2"),
     ],
+    args = ["build"],
```
NikolaMilosa, line 37: If you add this we won't be able to do the following 
```
bazel run //:mkdocs serve
```
Which is used for previewing docs locally

bazel/conf/.bazelrc.build
```diff
@@ -22,6 +22,10 @@ build --@rules_rust//:rustfmt.toml=//:rustfmt.toml
 # lockfile for MODULE.bazel
 common --lockfile_mode=off
 
+# Mypy support.
+# Targets with tag "typecheck" will get type-checked when built, run or tested.
+common --enable_bzlmod --aspects //tools:aspects.bzl%mypy_aspect --output_groups=+mypy
```
NikolaMilosa, line 27: `--enable_bzlmod` is a default, no need to specify it

release-controller/BUILD.bazel
```diff
@@ -51,10 +49,6 @@ py_binary(
     main = "commit_annotator.py",
     srcs = glob(
         ["*.py"],
-        exclude = [
-            "test_*.py",
-            "pytest.py",
-        ],
```
NikolaMilosa, line 57: Any particular reason why are we now including test files in commit annotator?

NikolaMilosa (commented):
Feels like we are trying everything we can to make python feel like something it isn't. How about we rewrite it in rust and the modules that we can't recreate we keep as python scripts? 

This is just absurd amount of changes to review IMHO, I reviewed just bazel stuff

BUILD.bazel
DFINITYManu: Fixed.

bazel/conf/.bazelrc.build
DFINITYManu: Will remove.

DFINITYManu:
@NikolaMilosa I would be in favor of rewriting at least the reconciler in Rust.  However, with these changes, it's far easier to work on the reconciler.

sasa-tomic (approved):
LGTM!

release-controller/dryrun.py
```diff
@@ -70,91 +46,115 @@ def __init__(self) -> None:
...
         post = self.post_by_id(post_id)
         assert post
-        if content != post["content"]:
+        if content != post["raw"]:
             self._logger.warning(
-                "Updating post %s titled %r",
+                "Updating post %s in topic %s",
                 post["id"],
-                post.get("title", "(no title)"),
+                post["topic_id"],
```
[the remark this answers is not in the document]
DFINITYManu: Yep we know because of typing.

image
image

Typing is the whole reason I was confident with the refactor.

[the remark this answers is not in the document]
sasa-tomic: Aha, great, thanks!

PR #1219, opened 2025-01-17
chore: migrate to bazel 7 and rust 1.84

NikolaMilosa:


MODULE.bazel
```diff
@@ -1,18 +1,129 @@
...
-#    requirements_txt = "//:requirements.lock",
-#)
-#use_repo(types, "pip_types")
+bazel_dep(name = "rules_pkg", version = "1.0.1")
+bazel_dep(name = "rules_oci", version = "2.0.0")
+
+oci = use_extension("@rules_oci//oci:extensions.bzl", "oci")
+oci.pull(
+    name = "bazel_image_6_5_0",
+    image = "gcr.io/bazel-public/bazel@sha256:7506002de2e90431e194f8391f4600b57cf229280002b2fc473b88a95d7104bc",
```
sasa-tomic, line 28: Should this also be bumped? I see there is a newer 7.x image at gcr.io/bazel-public/bazel

rs/cli/src/ic_admin.rs
```diff
@@ -18,6 +18,8 @@ use strum::Display;
 const MAX_SUMMARY_CHAR_COUNT: usize = 29000;
 
 #[automock]
+// If we allow this lint, automock complains
```
sasa-tomic, line 21:
```suggestion
// automock complains without the explicit allow below
```

sasa-tomic (approved):
TY!

MODULE.bazel
NikolaMilosa: That is a great catch. If we use bazel to do some queries in our `git` repo in release controller runtime this will break. I will bump it

PR #1223, opened 2025-01-17
chore: skipping all `chore(release)` commits from dre changelog

NikolaMilosa:
This PR will filter all `chore(release)` commits from `dre` tool changelog so that we keep it small


sasa-tomic (approved):
:rocket:

PR #1225, opened 2025-01-17
feat: offboard nodes in GE1 DC

sasa-tomic:
### Summary

- **Cordoned Features Update:**
  - Added multiple `node_id` entries in `cordoned_features.yaml` for offboarding the second rack of nodes in the GE1 DC, linking to relevant forum proposals for reference.

- **Code Enhancements in `rs/ic-management-types`:**
  - Added `NodeFeature::NodeId` to the node features serialization list to ensure inclusion.
  - Updated the ordering in `NodeFeature` enum and adjusted related logic to filter out `node_id` from specific processes.


NikolaMilosa (approved):
Would be nice in the future to add some unit tests for the cordoning functionalities.

PR #1240, opened 2025-01-21
fix(dre): dont require discourse secrets for actions

NikolaMilosa:
Previously the tool would error out if nothing is provided.

With this the tool will use placeholder key and user which will fail to create posts but the user will be prompted to create them manually and provide the links.


sasa-tomic (approved):
Thanks!

PR #1242, opened 2025-01-22
Get blessed HostOS versions using `dre registry`.

DFINITYManu:


release-controller/dre_cli.py
```diff
@@ -69,6 +69,21 @@ def get_blessed_versions(self) -> list[str]:
...
+        """Query the blessed versions."""
+        return set(
+            typing.cast(
+                list[str],
+                [
+                    n["hostos_version_id"]
+                    for n in json.loads(
+                        subprocess.check_output([self.cli, "registry"], env=self.env)
+                    )["nodes"]
+                    if "hostos_version_id" in n and n["hostos_version_id"].strip()
```
sasa-tomic, line 82:
```suggestion
                    if n.get("hostos_version_id", "").strip()
```

release-controller/dre_cli.py
```diff
@@ -69,6 +69,21 @@ def get_blessed_versions(self) -> list[str]:
             )["value"]["blessed_version_ids"],
         )
 
+    def get_blessed_hostos_versions(self) -> set[str]:
+        """Query the blessed versions."""
+        return set(
+            typing.cast(
+                list[str],
+                [
```
sasa-tomic, line 77: why not a set here?
```suggestion
                {
```

DFINITYManu: First I cast to list for typing, then I set.  There's basically no difference in performance at this scale..

sasa-tomic (approved):
cool

PR #1253, opened 2025-01-28
Implement the motion feature in DRE.

DFINITYManu:
User must specify a `--proposal-url` for discussion of the proposal, a single non-option argument specifying a file (`-` for stdin) to supply a proposal summary, and optionally a `--title` / a `--motion-text` / a `-motion-text-file`.

If not run with `--dry-run`, this command will not ask you for confirmation.  This is compatible with the way DRE does things (e.g. `dre vote`), and breaks away from the convention of `ic-admin`.  Run it with `--dry-run` and verify the proposal payload is what you meant it to be, before submitting.

Parameters as per right now:

image

### DIFF
--- a/rs/cli/src/commands/governance/propose/motion.rs
+++ b/rs/cli/src/commands/governance/propose/motion.rs
@@ -84,13 +84,9 @@ impl ExecutableCommand for Motion {
             println!("Proposal that would have been sent:\n{:#?}", proposal);
             return Ok(());
         }
-        match governance.make_proposal(NeuronId { id: neuron.neuron_id }, proposal.into()).await {
-            Ok(propresp) => {
-                println!("{:?}", propresp);
-                Ok(())
-            }
-            Err(e) => Err(e),
-        }
+        let propresp = governance.make_proposal(NeuronId { id: neuron.neuron_id }, proposal.into()).await?;
+        println!("{:?}", propresp);
+        Ok(())
     }
 
     fn validate(&self, _args: &crate::commands::Args, _cmd: &mut clap::Command) {}
--- a/rs/cli/src/commands/governance/propose/motion.rs
+++ b/rs/cli/src/commands/governance/propose/motion.rs
@@ -33,7 +33,7 @@ struct MotionParameters {
     #[arg(long, help_heading = "Command parameters", conflicts_with = "motion_text_file")]
     pub motion_text: Option<String>,
 
-    /// URL for discussion of the proposal; defaults to the DFINITY forum governance topic list if unspecified
+    /// URL typically used to discuss the proposal; defaults to the DFINITY forum governance topic list if unspecified
     #[arg(long, help_heading = "Command parameters", default_value = "https://forum.dfinity.org/c/governance/27")]
     pub proposal_url: url::Url,
 }
--- a/rs/cli/src/commands/governance/propose/motion.rs
+++ b/rs/cli/src/commands/governance/propose/motion.rs
@@ -65,7 +65,7 @@ impl Motion {
             // Strip out '#' plus any extra leading space
             let stripped_title = first_line.trim_start_matches('#').trim_start();
             let body = lines[1..].join("\n"); // Join the remaining lines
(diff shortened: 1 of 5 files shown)


rs/cli/src/commands/governance/propose/motion.rs
```diff
@@ -0,0 +1,97 @@
...
+    /// If no explicit --title is specified, the first headline found in the text will be stripped from the motion text.
+    #[clap(num_args(1..))]
+    pub motion_text_file: PathBuf,
+
+    /// Title to give to the proposal; defaults to the first Markdown level-1 heading of the motion text, which will be stripped from
+    /// the summary in the default case.  For this default to kick in, the heading must be at the top of the motion text.
+    #[clap(long)]
+    pub title: Option<String>,
+
+    // ATTENTION REVIEWERS: should we put the entire text of the proposal in the summary?  Or should we intelligently extract the
```
DFINITYManu, line 24: ATTENTION REVIEWERS.

pietrodimarco-dfinity: I believe always we should put the entire text of the proposal in the summary. Title should be probably passed in as arg.

DFINITYManu: We do.  But from what i have seen in proposals that exist, they do not repeat the title in the proposal.   That's why by default we mimic the same behavior.

rs/cli/src/commands/governance/propose/motion.rs
```diff
@@ -0,0 +1,93 @@
...
+    // first paragraph and use that?
+    /// Summary to give to the proposal; defaults to the text of the motion.
+    #[clap(long)]
+    pub summary: Option<String>,
+
+    /// URL for discussion of the proposal; defaults to no the DFINITY forum governance topic.
+    #[clap(long, default_value = "https://forum.dfinity.org/c/governance/27")]
```
sasa-tomic, line 31: With a wrong forum post link (which the default is), the motion proposal is certainly going to be rejected...

rs/cli/src/commands/governance/propose/motion.rs
```diff
@@ -0,0 +1,93 @@
...
+    pub url: url::Url,
```
sasa-tomic, line 32: elsewhere in the DRE tool this is a forum-post-link, and in ic-admin it's `--proposal-url`. I'd prefer if we don't introduce a 3rd term now.

rs/cli/src/commands/governance/propose/motion.rs
DFINITYManu: I think I have addressed this!

rs/cli/src/commands/governance/propose/motion.rs
DFINITYManu: I can fix this.

DFINITYManu: I'll call it `--proposal-url` since anyone can submit a proposal who may want to discuss that proposal in their own page / github / anything, doesn't have to be a forum or the DFINITY forum for that matter.

DFINITYManu: image

rs/cli/src/commands/governance/propose/motion.rs
DFINITYManu: Should we make it mandatory instead of a default?

DFINITYManu: image
image

DFINITYManu: I've made it mandatory.

DFINITYManu: Note the prior default was simply empty in the script that was used before: https://github.com/dfinity-ops/release/blob/main/scripts/motion-proposal/submit-motion-proposal.sh

sasa-tomic (approved):
TY!

PR #1262, opened 2025-01-31
Holistic Discourse / forum post link handling.

DFINITYManu:
* There are several modalities now: omit, ask, supply an URL, or use Discourse.
* Arguments for Discourse and forum post handling move to each individual command.
* When the Discourse behavior fails to create a post, it will ask for the user, *and* it will update the post afterwards.
* When a proposal fails, the user is reminded to delete any post that may have been created.
* When a proposal succeeds, and the link was manually supplied by the user, the user is reminded to update the post.
* Motion proposals have gained a formulaic forum post autocomposed based on the title and text of the proposal.  We should create more of these so as to avoid having to compose forum posts by hand.
* Formerly, new posts would be hidden by the update Discourse forum post call. Now, only posts submitted in reply to a topic are hidden after update.

To-do:

* When using the modality Discourse, if a proposal failed, then delete the corresponding forum post.  In principle this can be done easily for most calls (for all calls that wrap `ic_admin.propose_*`), but not all.


NikolaMilosa (approved):

sasa-tomic (approved):
I like it! :heart:

PR #1264, opened 2025-02-03
fix(reconciler): Use new subcommand of ic-admin for revising the GuestOS versions

DFINITYManu:
Additionally, we no longer place the proposal forum post link in the summary of the proposal. In its stead, we place the forum post link in the field of the proposal designated for it.

### DIFF
--- a/release-controller/dre_cli.py
+++ b/release-controller/dre_cli.py
@@ -47,14 +47,23 @@ class DRECli:
             ]
         else:
             self.auth = []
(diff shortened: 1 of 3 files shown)


NikolaMilosa (commented):
Do we still need to support the old proposal type? I highly doubt we will roll back to that. Also would be nice to use dre api for that `dre version revise-elected-versions guest-os ...` although it would require some tweaking to accept a payload from argument instead of prompting user in `$EDITOR`

PR #1278, opened 2025-02-10
Major refactor of proposal structures and processing of proposals.

DFINITYManu:
This is being submitted for review to gather questions and observations from you, because I am positive a refactor of this kind will raise many questions and I want to address them all before merging this.

Changes of note:

* ~`dre vote` no longer votes automatically by default, for consistency. Pass `--yes` to autovote.~
* This PR opens the door to a new kind of proposal (not based on IC admin) that will use the governance canister to directly submit proposals. That will allow us to delete some more code.  We now have the `ProposableViaIcAdmin` trait and the `ProposableViaGovernance` trait.
* The `forum_post_link` struct member has finally been deleted.  All forum post handling happens right before forrealz submission of a proposal, instead of the sui generis handling that took place before.


sasa-tomic (approved):
HUGE PR, making it very hard to review without spending a ton of time so let's just trust that you tested it thoroughly...

DFINITYManu:
> HUGE PR, making it very hard to review without spending a ton of time so let's just trust that you tested it thoroughly...

I'm sorry it got so big!  But all tests pass and I tested most each affected command.

PR #1279, opened 2025-02-10
feat(msd): adding possibility to remove stale targets

NikolaMilosa:
In case of testnets (be it farm testnets or new k8s testnets) we will be notified upon testnet creation but not on test removal. This PR aims to add the possibility for MSD to know which testnets are no longer active and remove them for the exported targets and scraping.

### DIFF
--- a/rs/ic-observability/multiservice-discovery/src/main.rs
+++ b/rs/ic-observability/multiservice-discovery/src/main.rs
@@ -5,6 +5,7 @@ use std::vec;
 
 use axum_otel_metrics::HttpMetricsLayerBuilder;
 use clap::Parser;
+use crossbeam_channel::{bounded, unbounded};
 use humantime::parse_duration;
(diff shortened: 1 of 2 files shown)


rs/ic-observability/multiservice-discovery/src/main.rs
```diff
@@ -260,4 +278,14 @@ exist, it will be created.
 "#
     )]
     networks_state_file: Option<PathBuf>,
+
+    #[clap(
+        long = "gc-timeout",
+        value_parser = parse_duration,
+        help = r#"
+If set, networks that cannot be scraped for `Duration` will be removed.
```
sasa-tomic, line 286:
```suggestion
If set, networks unreachable for at least the provided `Duration` will be automatically removed.
```

sasa-tomic (approved):
LGTM! TY!

PR #1290, opened 2025-02-19
Refactor: consolidate forum parameters and confirmation parameters into a single structure.

DFINITYManu:
What has happened here:

* Reduction of cyclic `use` dependencies.
* Auth stuff moved to `auth.rs`.
* `ExecutableCommand` stuff moved to `ctx` module, since (despite the name) it's just a vehicle to create a `Ctx` for commands to use.  I will revise this abstraction later on -- I think it may ultimately not be necessary.
* Consolidation of everything under `lib.rs` instead of double importing in `main.rs`.  Removes source of confusion and speeds up compilation.
* Confirmation options move to their own module `confirm.rs` to clear up confusion and the import graph.
* Forum submitter code from `forum.rs` moves to `submitter.rs` since it has dual responsibility (submit proposals and manage forum posts) so having it as a separate file adds clarity.
* Implementations of proposal executors from `proposal_executors.rs` move to `governance.rs` for the governance executor and `ic_admin.rs` for the ic-admin executor, breaking up an important import cycle.

One important change not in the list above:

The `SubmissionParameters` structure has sprung to existence to include the `ForumParameters` and `ConfirmationModeOptions` structs, and is now used when appropriate (in operations that require forum post parameters and can be affected by confirmation modes, e.g., dry run, confirm, unconditional).

These changes, together with prior changes that consolidate forum post management and proposal simulation / execution, make the code more maintainable, more explicit about what is implemented or happening and where it is, and pave the way for imbuing the `Proposable` types with knowledge about how to deal with the forum based on what the type of `Proposable`.

Noteworthy in light of the last sentence: now the Submitter type method `propose()` accepts a `ProposalExecution` (tied to a `Proposable`) next to a forum post type.  Soon, the type of forum post can be dictated by the `ProposalExecution` (effectively, the `Proposable`) itself, which makes sense, as we already have fixed forum post types for different proposals, but this knowledge is currently (incorrectly) hardcoded in an ad-hoc manner throughout various parts of the code.

**Important change: we hereby sunset the ability to pass `--yes`, `--no` and `--forum-post` stuff at arbitrary positions in the command line.  These are now accepted only after the subcommand name.  `clap` -- or perhaps our use of `clap` -- does not give us flexibility in parsing these flags anywhere in the command line.  I removed error-prone code that tried to guess (with only a moderate chance of success) under certain circumstances what the user meant.**

Find most current import graphs attached here:

modules

_Not all cycles have been eliminated!_


rs/cli/src/commands/api_boundary_nodes/add.rs
```diff
@@ -24,7 +22,7 @@ pub struct Add {
     pub motivation: Option<String>,
 
     #[clap(flatten)]
-    pub forum_parameters: ForumParameters,
+    pub submission_parameters: SubmissionParameters,
```
DFINITYManu, line 25: This change repeats everywhere.

rs/cli/src/commands/api_boundary_nodes/add.rs
```diff
@@ -33,23 +31,22 @@ impl ExecutableCommand for Add {
...
-                ic_admin::IcAdminProposalOptions {
-                    title: Some(format!("Add {} API boundary node(s)", self.nodes.len())),
-                    summary: Some(format!("Add {} API boundary node(s)", self.nodes.len())),
-                    motivation: self.motivation.clone(),
-                },
-            )),
-        )
-        .propose(ForumPostKind::Generic)
-        .await
+        Submitter::from(&self.submission_parameters)
```
DFINITYManu, line 34: This change also repeats everywhere.  Instead of explicitly passing the proposal execution, mode and forum stuff to method `from_executor_and_mode()`, now we turn the `SubmissionParameters` into a `Submitter`, then the execution itself is passed to the method `propose()` of the `Submitter`.

rs/cli/src/commands/mod.rs
```diff
@@ -148,14 +148,6 @@ Options:
     3. <commit> => specific commit"#)]
     pub ic_admin_version: IcAdminVersion,
 
-    /// To skip the confirmation prompt
-    #[clap(short, long, global = true, env = "YES", conflicts_with = "dry_run")]
-    pub yes: bool,
-
-    #[clap(long, aliases = [ "dry-run", "dryrun", "simulate", "no"], global = true, conflicts_with = "yes", help = r#"Dry-run, or simulate proposal submission. If specified will not submit the proposal
-but will show the ic-admin command and the proposal payload"#)]
```
DFINITYManu, line 156: All of this moves to the `forum` module where the submitter (which uses it)  lives.

rs/cli/src/commands/mod.rs
```diff
@@ -47,85 +45,7 @@ pub(crate) mod update_unassigned_nodes;
 pub mod upgrade;
 pub(crate) mod version;
 pub(crate) mod vote;
-
-/// HSM authentication parameters
-#[derive(ClapArgs, Debug, Clone)]
```
DFINITYManu, line 52: Stuff moves to auth.rs.

rs/cli/src/commands/propose.rs
```diff
@@ -1,77 +1,30 @@
...
     #[clap(flatten)]
-    pub forum_parameters: ForumParameters,
+    pub submission_parameters: SubmissionParameters,
 
     /// Arbitrary ic-admin args
     #[clap(allow_hyphen_values = true)]
     pub args: Vec<String>,
 }
 
-fn disambiguate_mode_from_ctx_and_postargs(mode: HowToProceed, postargs: &[String]) -> anyhow::Result<HowToProceed> {
```
DFINITYManu, line 21: Absolutely obsolete code.  We no longer support interspersing command parameters with ic-admin parameters.

rs/cli/src/confirm.rs
```diff
@@ -0,0 +1,52 @@
+use clap::Args as ClapArgs;
```
DFINITYManu, line 1: Confirmation options move to their own module.

sasa-tomic (approved):
LGTM, thanks for taking on this huge quality of life improvement!

PR #1300, opened 2025-02-24
feat: IC-1680: Add NPR V1 lib - Performance Multiplier Calculator

pietrodimarco-dfinity:
Node Provider Rewards V1 Library
This library computes performance-based rewards (in XDR) for node providers over a given reward period. Rewards are adjusted based on each node's performance using [trustworthy node metrics](https://internetcomputer.org/blog/features/node-metrics). The implementation follows the accepted proposal [135054](https://dashboard.internetcomputer.org/proposal/135054).

This PR implement the first step of the reward computation process:

- Add `PerformanceMultipliers` computation for each node based on its failure rate, derived from the number of blocks it failed to propose vs. successfully proposed.
- Includes tests for both the `PerformanceMultipliers` computation and library inputs.
- Adds detailed logging to track each step of the computation process.

Next Steps (TODO)
The next phase will apply the computed PerformanceMultipliers to the monthly XDR incentives each node receives for participating in the IC network.


rs/node-provider-rewards/src/lib.rs
```diff
@@ -0,0 +1,127 @@
...
+/// * reward_period - The time frame for which rewards are calculated.
+/// * metrics_by_node - Daily node metrics for nodes in `reward_period`. Only nodes in `providers_rewardable_nodes` keys are considered.
+/// * providers_rewardable_nodes: Nodes eligible for rewards, as recorded in the registry versions spanning the `reward_period` provided.
+///
+/// TODO: Implement the XDR reward calculation logic from the nodes multiplier.
+pub fn calculate_rewards(
+    reward_period: RewardPeriod,
+    metrics_by_node: BTreeMap<NodeId, Vec<NodeDailyMetrics>>,
+    providers_rewardable_nodes: BTreeMap<PrincipalId, Vec<NodeId>>,
+) -> Result<(), RewardCalculationError> {
```
DFINITYManu, line 30: `()` return type??

rs/node-provider-rewards/src/performance_calculator.rs
```diff
@@ -0,0 +1,250 @@
...
+impl PerformanceMultiplierCalculator {
+    pub fn new(
+        nodes_failure_rates: BTreeMap<NodeId, Vec<NodeDailyFailureRate>>,
+        subnets_failure_rates: BTreeMap<SubnetId, Vec<SubnetDailyFailureRate>>,
+    ) -> Self {
+        Self {
+            nodes_failure_rates,
+            subnets_failure_rates,
+            ctx: ExecutionContext {
+                execution_nodes: RefCell::new(BTreeMap::new()),
```
DFINITYManu, line 49: I'm not happy with the use of RefCell here.  It feels to me like trying to bypass the borrow checker.

pietrodimarco-dfinity: Ok will pass it in as mut to the different functions

DFINITYManu (approved):

PR #1301, opened 2025-02-24
Use proposal_id member for successful NNS proposal text, and handle DREMalfunction too

DFINITYManu:
Added logic to handle DREMalfunction in proposal retrieval, providing updated template for forum posts.


sasa-tomic (approved):
TY!

PR #1304, opened 2025-02-25
fix: not panicing for unknown operators because of low heights

NikolaMilosa:
For low heights it is possible that some structs will have missing data due to registry not having them at low heights.


sasa-tomic (approved):
TY!

PR #1305, opened 2025-02-25
chore: update `jira.js` and `linear@sdk`

NikolaMilosa:


NikolaMilosa:
Run it locally it worked... maybe we should unleash it and see what else breaks 😆

LittleChimera (approved):

PR #1307, opened 2025-02-25
feat: implement retry mechanism for Discourse API requests

sasa-tomic:
### Added
- Implemented a retry mechanism for sending requests to the Discourse API to handle `TOO_MANY_REQUESTS` errors by retrying with exponential backoff.

### Updated
- Modified existing request logic to encapsulate retry logic using a `for` loop.
- Updated error handling to include retries upon request failures, with a max retry count of 5 and incremental sleep before each retry.
- Included logging for rate limiting warnings and retry attempts.

### Removed
- Eliminated old match structure for direct request handling, replacing it with retry-inclusive logic.

### Misc
- Added unreachable!() statement to ensure code paths are handled explicitly.

### DIFF
--- a/rs/cli/src/forum/impls.rs
+++ b/rs/cli/src/forum/impls.rs
@@ -450,12 +450,12 @@ impl DiscourseClientImp {
                     }
                 }
                 Err(error) => {
-                    warn!("Error while sending request to Discourse: {}. Retrying in {} seconds", error, retry_secs);
-                    tokio::time::sleep(std::time::Duration::from_secs(retry_secs)).await;
-
                     if retry_num == retries_max {
                         return Err(DiscourseClientImpError::OtherReqwestError(error));
                     }
+
+                    warn!("Error while sending request to Discourse: {}. Retrying in {} seconds", error, retry_secs);
+                    tokio::time::sleep(std::time::Duration::from_secs(retry_secs)).await;
                 }
             }
         }
--- a/rs/cli/src/forum/impls.rs
+++ b/rs/cli/src/forum/impls.rs
@@ -419,25 +419,49 @@ impl DiscourseClientImp {
             request = request.body(payload);
         }
 
-        match request.send().await {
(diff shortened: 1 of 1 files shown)


rs/cli/src/forum/impls.rs
```diff
@@ -419,25 +419,49 @@ impl DiscourseClientImp {
...
+                            if response_status == reqwest::StatusCode::NOT_FOUND {
+                                return Err(DiscourseClientImpError::NotFound);
+                            }
+                            return Err(DiscourseClientImpError::OtherReqwestError(error));
                         }
                     }
                 }
+                Err(error) => {
+                    warn!("Error while sending request to Discourse: {}. Retrying in {} seconds", error, retry_secs);
+                    tokio::time::sleep(std::time::Duration::from_secs(retry_secs)).await;
```
DFINITYManu, line 454: dont sleep when retry num equals retry max

sasa-tomic: makes sense, fixed, PTAL

rs/cli/src/forum/impls.rs
```diff
@@ -419,25 +419,49 @@ impl DiscourseClientImp {
...
-                            Err(DiscourseClientImpError::OtherReqwestError(e))
+        let sleep_secs = 5;
+        let retries_max = 5;
+
+        for retry_num in 1..=retries_max {
+            let retry_secs = sleep_secs * retry_num;
+
+            match request.try_clone().unwrap().send().await {
+                Ok(response) => {
+                    if response.status() == reqwest::StatusCode::TOO_MANY_REQUESTS {
```
DFINITYManu, line 430: Should be within the error_for_status block below as another case

sasa-tomic: reasonable request, PTAL at the updated code

DFINITYManu (approved):

PR #1315, opened 2025-02-28
Reorder command line arguments in the DRE invocation.

DFINITYManu:
Moved the "--yes" flag after other arguments to comply with the fact that mode of operation arguments are part of the subcommand now.


sasa-tomic (approved):
LGTM

PR #1316, opened 2025-02-28
Update DRE CLI binary name, fixing brown paper bug.

DFINITYManu:
Changed the binary name from 'old-dre' to 'dre'.

This update ensures that the correct executable is used for DRE operations.


sasa-tomic (approved):
Makes sense!

PR #1318, opened 2025-02-28
Refactor DRE CLI propose command handling.

DFINITYManu:
Simplified propose command execution by removing redundant checks for obsolete ways of proposing GuestOS elections, and improving argument construction.

Updated reconciliation error logging to use exception for better traceability. We were swallowing the exception instead of printing it.


sasa-tomic (approved):
LGTM

PR #1319, opened 2025-02-28
The --yes must be immediately after the propose verb.

DFINITYManu:


release-controller/dre_cli.py
```diff
@@ -58,7 +72,6 @@ def _run(self, *args: str, **subprocess_kwargs: typing.Any) -> str:
                     self.cli,
                     *self.auth,
                     *args,
-                    *(["--yes"] if "propose" in args else []),
```
DFINITYManu, line 61: It's better not to try and divine what command the caller has sent.  The caller has better information on that :-D

sasa-tomic (approved):
Makes sense thanks

PR #1323, opened 2025-02-28
Print proposal ID when a proposal is issued.

DFINITYManu:
This PR refactors proposal submission handling on the Submitter to return `Option<ProposalResponseWithId>`.

With that refactor done, we can make a change across multiple CLI subcommands to ensure that the result of a successful proposal execution (as opposed to just simulation) is printed.

*Key changes:*

- Updated the `Submitter::propose` method signature to return `anyhow::Result<Option<ProposalResponseWithId>>`.
- Added detailed handling and logging for the proposal URL addition process in case of success or failure.
- Ensured CLI commands properly handle and display the proposal ID if returned by the submission process.

This adds compatibility with `ic-admin` to all proposals submitted with DRE, which in turn will be used to fix an issue in release controller.

### DIFF
--- a/rs/cli/src/commands/api_boundary_nodes/remove.rs
+++ b/rs/cli/src/commands/api_boundary_nodes/remove.rs
@@ -30,8 +30,8 @@ impl ExecutableCommand for Remove {
     }
 
     async fn execute(&self, ctx: crate::ctx::DreContext) -> anyhow::Result<()> {
-        if let Some(p) = Submitter::from(&self.submission_parameters)
-            .propose(
+        Submitter::from(&self.submission_parameters)
+            .propose_and_print(
                 ctx.ic_admin_executor().await?.execution(ic_admin::IcAdminProposal::new(
                     ic_admin::IcAdminProposalCommand::RemoveApiBoundaryNodes { nodes: self.nodes.to_vec() },
                     ic_admin::IcAdminProposalOptions {
@@ -42,11 +42,7 @@ impl ExecutableCommand for Remove {
                 )),
                 ForumPostKind::Generic,
             )
-            .await?
-        {
-            println!("{}", p)
-        };
-        Ok(())
+            .await
     }
 
     fn validate(&self, _args: &GlobalArgs, _cmd: &mut clap::Command) {}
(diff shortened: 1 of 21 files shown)


rs/cli/src/commands/api_boundary_nodes/remove.rs
```diff
@@ -42,7 +42,11 @@ impl ExecutableCommand for Remove {
                 )),
                 ForumPostKind::Generic,
             )
-            .await
+            .await?
+        {
+            println!("{}", p)
+        };
+        Ok(())
```
NikolaMilosa, line 49: A lot of these lines are duplicated across the codebase. Could we maybe add an option to submitter that would be something like: `emit_proposal_id: bool` and if its set in the `propose` method it would print?

DFINITYManu: Yes that was my first intuition.

But:

* That would add an additional side effect to the function, and I want to keep side effects minimized (at least, without this one additional side effect).
* It's conceivable we would like to call that function but only get the proposal ID back, not print it.  That code has not been written but might be (or, actually, I'd argue some of the code we have *already should behave that way* -- when we issue multiple proposals, for example -- but we can fix that later).
* The job of _deciding what to do with the result_ belongs in the layer of the subcommand, not the layer of the submitter.

That's why I did the tedious work of putting that same `println!()` everywhere.

NikolaMilosa: You could change it to a log, and then emit it always. Anyway, its not a blocker, this is now just not centralized, and changing this code again in the future will be tedious as all new proposals will have to remember to add a print

DFINITYManu: I have an idea.  I can create a separate Submitter method that does it for us, and then we can fold those code changes back.

DFINITYManu: Yeah I have implemented the change.  Does it look better now?

NikolaMilosa (approved):

PR #1338, opened 2025-03-06
feat(decentralization): add provider clusters for node linkage analysis

sasa-tomic:
### Added
- **provider_clusters.rs**: Initialized a map of linked providers.

### Updated
- **network.rs**:
  - Integrated `get_linked_providers` to assess provider linkages.
  - Implemented a scoring mechanism to penalize multiple nodes from linked providers within a subnet.

### Functionality
- Enhanced `DecentralizedSubnet` to evaluate and report on node distribution across potentially linked providers applying solution penalties where necessary.


rs/decentralization/src/provider_clusters.rs
```diff
@@ -0,0 +1,31 @@
+use ahash::{AHashMap, AHashSet};
+use ic_base_types::PrincipalId;
+use lazy_static::lazy_static;
+use std::str::FromStr;
+
+lazy_static! {
+    static ref POSSIBLY_LINKED_PROVIDERS: AHashMap<String, AHashSet<PrincipalId>> = AHashMap::from_iter([
```
DFINITYManu, line 7: I'm in principle OK with this change, but there is nothing in the code that explains why it's necessary, with some form of backing that's more substantial than someone's opinion.

Let's fix this please.

DFINITYManu (changes requested):

rs/decentralization/src/provider_clusters.rs
sasa-tomic: It was agreed to remove this cluster of providers, and for the other cluster of providers that was self-explained in the forum, to add a link to the forum post where it was explained

DFINITYManu (approved):

PR #1355, opened 2025-03-11
docs: op-connect-server certificate issue

NikolaMilosa:
Adds the recent debug steps for op secrets store being out of sync

### DIFF
--- a/docs/k8s/tips-and-tricks.md
+++ b/docs/k8s/tips-and-tricks.md
@@ -12,3 +12,74 @@
     ```
     kubectl -n flux-system scale deployment kustomize-controller --replicas 1 --as root
     ```
+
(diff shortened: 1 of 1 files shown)


sasa-tomic (approved):
Awesome, thanks!

LittleChimera:
this needs a permanent fix instead

LittleChimera (changes requested):
this needs a permanent fix instead

DFINITYManu (changes requested):
is this a workaround?  do we have a runbook for this alert?  if so, this should probably be included in the runbook.

NikolaMilosa:
Discussed: https://dfinity.slack.com/archives/C03CE8KMM7V/p1741695903680589

PR #1387, opened 2025-03-17
A twofer: fix ugly error message in annotator, and enhance docs.

DFINITYManu:
The first thing in this commit is to remove this message from the standard error:

```
error: no note found for object 7d4e057cbb17f4bf2a108d547402cb5e7bf3eacc.
```

which is unnnecessary.  I had even added the functionality to do this very thing into the `notes()` function before, but forgot to call it when it was appropriate.  The code does the right thing -- but it just looks ugly when doing it.

The second thing is the URL to the Jira ops board which should be fixed to exclude closed alerts because they are noise and should not be shown by default.  So we I it here.

I'll handle rollout of the annotator after this.

### DIFF
--- a/docs/be-dr-dre.md
+++ b/docs/be-dr-dre.md
@@ -43,8 +43,8 @@ The release process [is documented here](https://www.notion.so/dfinityorg/IC-OS-
 
 ### 2. **Review alerts for our clusters**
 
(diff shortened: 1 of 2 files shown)


sasa-tomic (approved):
TY!

PR #1388, opened 2025-03-18
feat: IC-1680: Add NPR V1 lib - Rewards Calculation

pietrodimarco-dfinity:
**Node Provider Rewards V1 Library**
Second phase of NPR_V1 implementation [here](https://github.com/dfinity/dre/pull/1300) the first phase.

In this phase the computed `PerformanceMultipliers` are applied to the monthly XDR Rewards each node receives for participating in the IC network.

- Add `RewardsCalculator` which computes rewards per Node Provider given nodes `PerformanceMultipliers` and `NodeRewardsTable`
- Extend `ExecutionContext` functionality used now as holder for final/intermediate results
- The entire computation is now tracked in a `ComputationTable`
- Change the way intermediate results are logged now represented in table columns.

Next Steps (TODO)
- Add more tests focused on the `RewardsCalculator`
- Add one more table representing the computation of `BaseRewards` for each `NodeCategory`
- Store `IntermediateResults` in stable memory


DFINITYManu (commented):
A small refactor is taking place, I will review again after the refactor.  Code is looking quite good!

rs/node-provider-rewards/src/lib.rs
```diff
@@ -1,61 +1,80 @@
...
-    let perf_calculator = PerformanceMultiplierCalculator::new(nodes_failure_rates, subnets_failure_rates);
+    for (provider_id, provider_nodes) in rewardable_nodes_by_provider(rewardable_nodes) {
+        let RewardsCalculationResult {
+            rewards,
+            computation_log_tabled,
+        } = ctx.calculate_rewards(provider_nodes);
 
-    for (_provider_id, provider_nodes) in providers_rewardable_nodes {
-        let _nodes_multiplier = perf_calculator.calculate_performance_multipliers(&provider_nodes);
+        rewards_per_provider.insert(provider_id, rewards.to_u64().expect("Conversion succeeded"));
```
DFINITYManu, line 56: Please deal with the error instead of relying on expect, preferably by using the correct data type in the source datum.

rs/node-provider-rewards/src/types.rs
```diff
@@ -0,0 +1,34 @@
...
+impl RewardableNode {
+    pub fn category(&self) -> NodeCategory {
+        NodeCategory {
+            region: self.region.clone(),
+            node_type: self.node_type.clone(),
+        }
+    }
+}
+
+pub fn rewardable_nodes_by_provider(rewardable_nodes: &[RewardableNode]) -> BTreeMap<PrincipalId, Vec<RewardableNode>> {
```
DFINITYManu, line 27: Consider sort of inbound iter then https://docs.rs/itertools/0.7.4/itertools/trait.Itertools.html#method.group_by to produce the result you want.

DFINITYManu (approved):

PR #1389, opened 2025-03-19
feat: Add Prometheus telemetry to node provider rewards canister.

DFINITYManu:
This commit adds very basic support for Prometheus telemetry (including a mechanism to encode metrics using the `ic-metrics-encoder` crate).

The metrics added are:

* Generic memory usage coming from IC NNS metrics code.
* Metrics calculation start, success and end timestamps.


rs/dre-canisters/node-provider-rewards/src/lib.rs
```diff
@@ -19,21 +23,56 @@ pub type RegistryStoreInstance = CanisterRegistryStore<State, VM>;
 pub type MetricsManagerInstance = MetricsManager<State, VM>;
 pub const IC_CANISTER_CLIENT: ICCanisterClient = ICCanisterClient {};
 
+#[derive(Default)]
+pub struct PrometheusMetrics {
+    last_calculation_start: f64,
+    last_calculation_success: f64,
+    last_calculation_end: f64,
```
sasa-tomic, line 30: Are we planning to add more metrics later?

rs/dre-canisters/node-provider-rewards/src/lib.rs
```diff
@@ -19,21 +23,56 @@ pub type RegistryStoreInstance = CanisterRegistryStore<State, VM>;
...
+
+    fn mark_last_calculation_success(&mut self) {
+        self.last_calculation_end = (ic_cdk::api::time() / 1_000_000_000) as f64;
+        self.last_calculation_success = self.last_calculation_end
+    }
+
+    fn mark_last_calculation_end(&mut self) {
+        self.last_calculation_end = (ic_cdk::api::time() / 1_000_000_000) as f64
+    }
+}
```
sasa-tomic, line 50: WDYT about just adding a method for serving metrics instead of relying on an external crate for it? Something like:
```
fn serve_metrics() -> HttpResponse {
    match metrics() {
        Ok(body) => {
            let headers = vec![
                (
                    "Content-Type".to_string(),
                    "text/plain; version=0.0.4".to_string(),
                ),
                ("Content-Length".to_string(), body.len().to_string()),
            ];
            HttpResponse {
                status_code: 200,
                headers,
                body: ByteBuf::from(body),
                upgrade: None,
            }
        }
        Err(err) => HttpResponse {
            status_code: 500,
            headers: vec![],
            body: ByteBuf::from(format!("Failed to encode metrics: {err}")),
            upgrade: None,
        },
    }
}
```

rs/dre-canisters/node-provider-rewards/src/lib.rs
DFINITYManu: It's likely.  Depends on what we want to find out, which in turn depends on what the calculator looks like in the end.

rs/dre-canisters/node-provider-rewards/src/lib.rs
DFINITYManu: I would prefer to stick to current practice here.

DFINITYManu: We're doing the same as the governance canister.  It's also convenient that I can just call a function call encode_metric for each metric, instead of having to roll my own text encoding here.

sasa-tomic (approved):

PR #1412, opened 2025-03-25
Path checking for multiple glob patterns, and container rollout link in job summary.

DFINITYManu:
This update allows the action to handle space-separated lists of bash [[]]-compatible globs for path checking.

The modified script iterates over each file in the diff and checks it against each pattern, updating the output accordingly.

Globbing is turned off so the second-nested for loop won't get globbing to interfere with it.

As a further enhancement, now the workflow step checks for a diff range rather than just the last commit.  This makes the file modification checker work in many more and varied scenarios.

Finally, the Github workflow job summary feature looks like this:

image

### DIFF
--- a/.github/workflows/check-modified-files-as-step/action.yaml
+++ b/.github/workflows/check-modified-files-as-step/action.yaml
@@ -3,8 +3,12 @@ description: Reusable action for checking if a file path is modified as a step. 
 
 inputs:
   path:
-    description: "Path to check"
+    description: "Path to check -- can be a space-separated list of bash [[]]-compatible globs"
     required: true
+outputs:
+  changed:
+    description: "Whether any of the files matching the globs specified changed between the base and the head commit"
+    value: ${{ steps.check-changed.outputs.changed }}
 
 runs:
   using: composite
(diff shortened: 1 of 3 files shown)


.github/workflows/check-modified-files-as-step/action.yaml
```diff
@@ -3,35 +3,50 @@ description: Reusable action for checking if a file path is modified as a step.
...
     ########################################
-    - name: "🔍 Check if $PATH_TO_CHECK is modified in last commit "
+    - name: "🔍 Check whether $PATH_TO_CHECK were modified in between the commit range"
+      id: check-changed
       env:
         PATH_TO_CHECK: ${{ inputs.path }}
       shell: bash
       run: |
-        set -eExou pipefail
+        set -eEou pipefail
```
sasa-tomic, line 25: why dropping the -x globally? It's often helpful for debugging. We could do `set +x` in specific parts instead...

DFINITYManu: because this doesn't need debugging anymore, I debugged it until 11 pm. heh.

DFINITYManu: (the bash -x noise is a psychological price we should not pay unless necessary..)

PR #1428, opened 2025-03-27
Separate the release controller pipeline from the other pipelines.

DFINITYManu:
Should yield good resources saving when changes are made that do not involve the release controller.

This PR sits atop this other one: https://github.com/dfinity/dre/pull/1425

EDIT: 2 minutes is _not too shabby_:

image

EDIT 2:

Before:

image

After:

image

EDIT 3:

Improvements in Bazel BUILD files yielded this amazing run:

image


.github/workflows/main.yaml
```diff
@@ -6,11 +6,13 @@ on:
     paths-ignore:
       - "release-index.yaml"
       - "replica-releases/**"
+      - "release-controller/**"
```
sasa-tomic, line 9: should we also add a dependency for the python requirements - a change in the lock files could break it.

sasa-tomic (approved):

PR #1444, opened 2025-04-02
Make DRE CLI tests cacheable.

DFINITYManu:
Each time a push happens to a PR, `dre` gets rebuilt, because the version changes.  CLI unit tests depend on this target.  As a result, the tests are run all over again, wastefully, for no good reason at all.

This PR fixes that.

This change does *not* make the build of the `dre` target cacheable!  This change only makes the tests depend on the cacheable version of the target.

Before:

image

After:

image

Rant:

We should strongly consider not adding any version junk into any of our binaries **unless those binaries are going to go through an actual release pipeline** (e.g. the actual `dre` binary built for user download).  Stuff shipped in containers _does not need a version number!_  The container already has that information in the tag.

### DIFF
--- a/rs/cli/BUILD.bazel
+++ b/rs/cli/BUILD.bazel
@@ -42,7 +42,8 @@ rust_binary(
 )
 
 rust_binary(
-    name = "dre-embedded", # Variant does not have an embedded git revision, which permits caching of dependents.
+    # Variant does not have an embedded git revision, which permits caching of dependents.
+    name = "dre-embedded",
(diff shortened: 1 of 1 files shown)


DFINITYManu:
We should strongly consider not adding any version junk into any of our binaries **unless those binaries are going to go through an actual release pipeline**.  Stuff shipped in containers _does not need a version number!_  The container already has that information in the tag.

sasa-tomic (approved):
Thanks!

PR #1449, opened 2025-04-04
Un-rot the HostOS release notes.

DFINITYManu:
To be able to roll out HostOS soon, we have to un-rot this code.

This *really* should be replaced with an algorithm built into the release controller.  And the mechanism used in release  controller to detect file changes should be audited again, because if it's anything like this code, it's just wrong.

### DIFF
--- a/scripts/host-os-release-notes.py
+++ b/scripts/host-os-release-notes.py
@@ -1,8 +1,14 @@
 #!/usr/bin/env python3
+
+# FIXME: this script should be retired in favor of logic within the reconciler
+# that expands the work it does from just GuestOS to HostOS as well.
+
 import argparse
 import fnmatch
+import hashlib
 import os
 import pathlib
+import pickle
 import re
 import subprocess
 import sys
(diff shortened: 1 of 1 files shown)


scripts/host-os-release-notes.py
```diff
@@ -457,7 +528,7 @@ def main():
 
         change_infos[commit_type].append(
             {
-                "commit": merge_commit,
```
DFINITYManu, line 460: hey @sasa-tomic this  is why we were getting the wrong commit title / link.

DFINITYManu: like, whaaat

scripts/host-os-release-notes.py
```diff
@@ -139,19 +146,50 @@ def get_rc_branch(repo_dir, commit_hash):
...
-    rc_branches = [branch for branch in all_branches if branch.startswith("origin/rc--20")]
+    rc_branches = [
+        branch for branch in all_branches if branch.startswith("origin/rc--20")
+    ]
     if rc_branches:
         return rc_branches[0]
     return ""
 
 
+def memoize(f):
```
DFINITYManu, line 157: grog make slow thing go fast second time

PR #1474, opened 2025-04-15
Fix creation of rollout PRs for containers in k8s

DFINITYManu:
For some reason now Bazel is not actually printing when it's pushing containers out to the container registry.  This broke the creation of PRs to update containers in k8s.

Thus, we now unconditionally create or update the rollout PRs when merging to `main`, and additionally when the workflow is manually dispatched.


sasa-tomic (approved):
LGTM, thanks

PR #1479, opened 2025-04-16
Refactor PrometheusMetrics to separate telemetry module.

DFINITYManu:
Moved PrometheusMetrics from lib.rs to its own telemetry module for better organization and separation of concerns.


DFINITYManu:
Latest commit contains telemetry for the instructions but it's basically worthless because we're not actually counting the instructions of the most heavyweight functions since the instruction counter effectively may or may not reset when `.await`.

DFINITYManu:
I have deleted the commit with the instruction counter metrics.  It's pointless and doesn't belong in this refactor.

pietrodimarco-dfinity (approved):

PR #1480, opened 2025-04-16
Instruction counter metrics.

DFINITYManu:
This PR introduces instruction counters (as Prometheus gauges) that allow telemetry observers to see:

* the count of instructions that the last sync incurred,
* specific counts of instructions for subnet list, update metrics, registry sync, specifically during the last sync.

### DIFF
--- a/rs/dre-canisters/node-provider-rewards/canister/src/lib.rs
+++ b/rs/dre-canisters/node-provider-rewards/canister/src/lib.rs
@@ -18,20 +18,44 @@ mod telemetry;
 const HOUR_IN_SECONDS: u64 = 60 * 60;
 const DAY_IN_SECONDS: u64 = HOUR_IN_SECONDS * 24;
 
+struct InstructionCounter {
+    start: u64,
+    total: u64,
+}
+
+impl InstructionCounter {
+    fn new() -> Self {
+        Self {
+            start: ic_cdk::api::instruction_counter(),
+            total: 0,
+        }
+    }
+    fn add_up(&mut self) -> u64 {
+        let now = ic_cdk::api::instruction_counter();
+        self.start = now;
+        self.total += now;
+        self.total
+    }
+}
+
 /// Sync the local registry and subnets metrics with remote
 ///
 /// - Sync local registry stored from the remote registry canister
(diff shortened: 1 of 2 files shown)


rs/dre-canisters/node-provider-rewards/canister/src/lib.rs
```diff
@@ -40,6 +51,13 @@ async fn sync_all() {
             ic_cdk::println!("Failed to sync local registry: {:?}", e)
         }
     }
+
+    telemetry::PROMETHEUS_METRICS.with_borrow_mut(|m| {
```
pietrodimarco-dfinity, line 55: I would move this in the METRICS_MANAGER and REGISTRY_STORE so that they are in charge of updating their own metrics. 
We would need additional metrics anyway that are computed inside those instances

pietrodimarco-dfinity (changes requested):

rs/dre-canisters/node-provider-rewards/canister/src/lib.rs
DFINITYManu: I wouldn't -- because these metrics are metrics specifically _for the last sync_, not for when those objects' methods are called.

pietrodimarco-dfinity: But I mean the design of these objects shouldn't care if these metrics are called from an update context or whatever they should just update their own metrics. This will also become more and more complex as we add metrics.
Say that we want now as metric the last local registry version stored for example...
Btw not a problem :) I will approve we can change later in case we need

rs/dre-canisters/node-provider-rewards/canister/src/lib.rs
```diff
@@ -23,15 +23,26 @@ const DAY_IN_SECONDS: u64 = HOUR_IN_SECONDS * 24;
 /// - Sync local registry stored from the remote registry canister
 /// - Sync subnets metrics from the management canister of the different subnets
 async fn sync_all() {
+    let mut counter = telemetry::InstructionCounter::new();
```
pietrodimarco-dfinity, line 26: maybe instruction_counter is more appropriate?

PR #1487, opened 2025-04-17
fix: offset in log fetcher

NikolaMilosa:
https://www.freedesktop.org/software/systemd/man/latest/systemd-journal-gatewayd.html#Range%20header


sasa-tomic (approved):
LGTM

PR #1490, opened 2025-04-17
fix(reconciler): Wrap the entire phase of release notes preparation into the phase decorator.

DFINITYManu:
With this, we will get telemetry of successful release notes preparation even when the notes do not need to be prepared, enhancing the consistency of the telemetry we get.

This also allows us to capture and record telemetry for any exceptions within the check portion of the phase, which was previously not wrapped in the phase decorator.

### DIFF
--- a/release-controller/reconciler.py
+++ b/release-controller/reconciler.py
@@ -498,41 +498,43 @@ class Reconciler:
                 rclogger.debug("Ensuring forum post for release candidate exists.")
                 rc_forum_topic = self.forum_client.get_or_create(v.rc)
 
-            if markdown_file := self.notes_client.markdown_file(
-                release_commit, v.os_kind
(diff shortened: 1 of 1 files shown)


release-controller/reconciler.py
```diff
@@ -498,41 +498,43 @@ def reconcile(self) -> None:
                 rclogger.debug("Ensuring forum post for release candidate exists.")
                 rc_forum_topic = self.forum_client.get_or_create(v.rc)
 
-            if markdown_file := self.notes_client.markdown_file(
```
DFINITYManu, line 501: Disable whitespace in the diff to see the change better!

sasa-tomic (approved):
Nice!

LittleChimera (approved):

PR #1531, opened 2025-05-07
feat(node-provider-rewards): Perform inter-canister calls using non-blocking calls

pietrodimarco-dfinity:
Changes:

- Bump `ic-cdk `dep to 0.18 (where bounded calls are supported)
- Adapt metrics fetching to use bounded calls
- Adapt tests 
- Introducing `RegistryClient` to query registry using bounded calls

TODO:

- Remove `RegistryClient` once `RegistryClient` in IC repo depends on `ic-cdk 0.18`

### DIFF
--- a/rs/dre-canisters/node-provider-rewards/canister/src/lib.rs
+++ b/rs/dre-canisters/node-provider-rewards/canister/src/lib.rs
@@ -194,7 +194,7 @@ fn post_upgrade() {
     setup_timers();
 }
 
-#[query]
+#[query(hidden = true)]
 fn http_request(request: HttpRequest) -> HttpResponse {
     match request.path() {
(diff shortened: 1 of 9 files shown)


rs/dre-canisters/node-provider-rewards/canister/src/lib.rs
```diff
@@ -193,7 +194,7 @@ fn post_upgrade() {
     setup_timers();
 }
 
-#[query(hidden = true, decoding_quota = 10000)]
```
NikolaMilosa, line 196: Doesn't this mean that, if the candid generator hits it, it will try and add it to the `services` part and fail due to `HttpRequest` and `HttpResponse` not being a Candid type?

DFINITYManu (approved):

PR #1533, opened 2025-05-07
Manual annotation mode for commit annotator.

DFINITYManu:
You can now run:

```bash
bazel run //release-controller:commit-annotator -- manually-annotate COMMIT_ID yes/no [--os-kind GuestOS/HostOS]
```

to manually annotate a commit as affecting a specific OS kind (or both, if left out).

Useful to help the commit annotator "get over a hump" when a commit pushed to the IC repo cannot be built by Bazel.


NikolaMilosa (approved):
Nooooice! Thanks

PR #1556, opened 2025-05-21
Add additional info to registry-dump and changes node_allowance_total, rewards_correct

pietrodimarco-dfinity:


rs/cli/src/commands/registry.rs
```diff
@@ -195,13 +190,17 @@ async fn get_node_operators(local_registry: &Arc<dyn LazyRegistry>, network: &Ne
...
-            // Find the number of nodes registered by this operator
-            let operator_registered_nodes_num = all_nodes.iter().filter(|(nk, _)| nk == &k).count() as u64;
             let nodes_in_subnets = all_nodes
                 .iter()
                 .filter(|(_, value)| value.operator.principal == record.principal && value.subnet_id.is_some())
                 .count() as u64;
             let nodes_in_registry = all_nodes.iter().filter(|(_, value)| value.operator.principal == record.principal).count() as u64;
+            let node_allowance_total = record.node_allowance + nodes_in_registry;
+            let nodes_rewards_types_count = record.rewardable_nodes.values().map(|count| *count as u64).sum::<u64>();
+            let all_types_present = nodes_rewards_types_count == node_allowance_total;
```
sasa-tomic, line 200: Should this be something like `all_nodes_present_and_rewarded`? The current `all_types_present` isn't very clear.

rs/cli/src/commands/registry.rs
```diff
@@ -195,13 +190,17 @@ async fn get_node_operators(local_registry: &Arc<dyn LazyRegistry>, network: &Ne
...
+            let rewards_correct = nodes_rewards_types_count == nodes_in_registry;
+            let more_than_one_type = record.rewardable_nodes.len() > 1;
```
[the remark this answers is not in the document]
pietrodimarco-dfinity: Yeah I wanted to consider with this also the ones with 0 count to be used as input for @alexandru-uta to see if this type (even the one with 0 count) should be consider for creating `max_rewardable_count`

rs/cli/src/commands/registry.rs
pietrodimarco-dfinity: currently until we don't want to even consider delivering rewards until all nodes are strictly typed so I wouldn't go with that but what about `all_nodes_present_and_typed`?

sasa-tomic: Sure, SGTM!

rs/cli/src/commands/registry.rs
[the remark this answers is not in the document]
sasa-tomic: I miss context for the `max_rewardable_count` - what is that?

PR #1564, opened 2025-05-23
Release controller fixes

DFINITYManu:
* Process for proposal creation only latest two release.
* Handle potential errors fetching the image packages better.
* Allow override of the base version used for the changelog in the index.
* Consult dashboard API (optionally) to obtain latest proposals, and thus prevent resubmission of versions already submitted in the past.  The option to disable this behavior has been removed.

### DIFF
--- a/rs/cli/src/commands/proposals/filter.rs
+++ b/rs/cli/src/commands/proposals/filter.rs
@@ -284,6 +284,7 @@ impl ExecutableCommand for Filter {
                 include_all_manage_neuron_proposals: Some(true),
                 ..Default::default()
             };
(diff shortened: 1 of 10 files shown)


rs/cli/src/commands/proposals/filter.rs
```diff
@@ -284,6 +284,7 @@ impl ExecutableCommand for Filter {
                 include_all_manage_neuron_proposals: Some(true),
                 ..Default::default()
             };
+            info!("{:?}", payload);
```
sasa-tomic, line 287: leftover?

sasa-tomic (approved):

PR #1582, opened 2025-06-10
Improve logging in the reconciler

DFINITYManu:
* Add package checksum and unelected revisions logging, to verify during the next release that revisions are being unelected.
* Remove constant "Running version 0.0.0" when running `dre-embedded` in the reconciler.  Pointless, spams the log.

### DIFF
--- a/rs/cli/src/main.rs
+++ b/rs/cli/src/main.rs
@@ -13,7 +13,9 @@ use log::{info, warn};
 async fn main() -> anyhow::Result<()> {
     init_logger();
     let curr_version = env!("CARGO_PKG_VERSION");
-    info!("Running version {}", curr_version);
+    if curr_version != "0.0.0" {
+        info!("Running version {}", curr_version);
+    }
 
     dotenv().ok();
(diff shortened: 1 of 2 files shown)


NikolaMilosa (approved):

rs/cli/src/main.rs
```diff
@@ -13,7 +13,9 @@ use log::{info, warn};
 async fn main() -> anyhow::Result<()> {
     init_logger();
     let curr_version = env!("CARGO_PKG_VERSION");
-    info!("Running version {}", curr_version);
+    if curr_version != "0.0.0" {
+        info!("Running version {}", curr_version);
+    }
```
NikolaMilosa, line 18: Will be weird to not see the version even if it is this one. Any specific reason as to why we skip printing it?

DFINITYManu: yeah it's what prints now

image

that's what the release controller prints every time the DRE tool is invoked (every 60 seconds or so).

PR #1584, opened 2025-06-11
Fix(dre): Handle unexpected zero node reward types count

DFINITYManu:
Resolved panic by logging warning and returning an empty map when the node operator has nodes in the registry and rewardable nodes, but somehow no node reward types.

- Replaced `panic!` with `warn!` to log the issue instead of terminating execution.
- Provided detailed information in the log, including the principal, total nodes allowed, and reward types count.

Addresses (but does not necessarily fix) https://dfinity.atlassian.net/browse/DRE-466 .


NikolaMilosa (approved):
I'll approve for now but this should be reviewed by @pietrodimarco-dfinity

PR #1585, opened 2025-06-11
Disable team reviews for changelogs of HostOS and GuestOS

DFINITYManu:
With this, Slack announcements will continue to happen when release notes are prepared for review.  However, no teams will be pinged on Slack or listed in the document for cross-out to mark review as reviewed. This in turn means that the reconciler will proceed immediately to publishing the changelog as-is to the Github repository, in preparation for creation of the proposal.

Motivated by teams asking not to be reviewing the changelog anymore.

We elect to keep the functionality to have the option of re-enabling changelog review, should the future show us that a team wants to review their changes again.

Closes https://dfinity.atlassian.net/browse/DRE-465 .

### DIFF
--- a/release-controller/publish_notes.py
+++ b/release-controller/publish_notes.py
@@ -169,6 +169,12 @@ class PublishNotesClient:
                 "Could not find release notes section for version %s" % version
             )
 
+        # Attempt to find NO text between the Review checklist sentence and the Release notes headline.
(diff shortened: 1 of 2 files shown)


DFINITYManu:
Fixing tests...

LittleChimera (approved):

PR #1674, opened 2025-08-13
chore: adding guest launch measurements to the proposal

NikolaMilosa:
Soon, this will become mandatory. 

This PR can wait for @DFINITYManu

### DIFF
--- a/release-controller/dre_cli.py
+++ b/release-controller/dre_cli.py
@@ -1,4 +1,5 @@
 import json
+import sys
 import logging
 import subprocess
 import typing
@@ -239,6 +240,7 @@ class DRECli:
         else:
             # We will not parse the text here.  We dry-ran the thing, after all,
             # so there will be no proposal ID to parse.
+            print(text, file=sys.stderr)
             return 0
 
 


release-controller/dryrun.py
```diff
@@ -285,15 +287,29 @@ def announce_release(
...
+        f.flush()
+        dre.propose_to_revise_elected_os_versions(
+            changelog=changelog,
+            version="0" * 40,
+            os_kind=GUESTOS,
+            forum_post_url="https://forum.dfinity.org/t/proposal-to-elect-new-release-rc-2024-03-27-23-01/29042/7",
+            unelect_versions=[],
+            package_checksum="0" * 40,
+            package_urls=["https://doesntmatter.com/"],
+            launch_measurements_path=pathlib.Path(f.name),
```
DFINITYManu, line 311: Good change, but the API in this function should **not** take a path.

It should take an array of bytes, and the string should be written inside the function to a temporary file.  IOW the parameter ought to be `Optional[bytes]`.  That way we do not litter the callers with calls to file.write.  To get _bytes_ from the response, instead of using `response.text`,  use `response.content`.

Also no hardcoded paths please.

DFINITYManu (changes requested):

DFINITYManu (changes requested):
x

DFINITYManu (approved):

PR #1699, opened 2025-09-01
feat(subnet): rename include -> add-nodes and validate node availability

sasa-tomic:
### Motivation
- Fail fast when user requests specific nodes that are unavailable/unhealthy.
### Solution
- Rename CLI/requests/traits field include -> add_nodes and add --add-nodes (with aliases); use adding_from_available.
- Add validate_nodes_available_and_healthy and early checks in Runner/SubnetManager; aggregate errors.
### Impact
- Breaking: callers must use add_nodes/--add-nodes instead of include/--include.
### Meta — updated tests accordingly


NikolaMilosa (approved):
Aweeeesomeeeee 💯

PR #1725, opened 2025-09-18
chore(release): Elect version 51654f72ddf859624c0e2c19823c74104c9f983c as HostOS candidate for rollout

sa-github-api:


NikolaMilosa:
This version has a bug and we decided to not merge it

PR #1754, opened 2025-10-07
Stop release controller from updating all Discourse posts on its firs…

DFINITYManu:
…t after-boot cycle.

Analysis of what release controller was doing (thanks to the post diff feature added yesterday) led us to discover that posts were generated with a carriage return at the end, which Discourse happily accepts but then promptly chomps.  Thus, every time the code tried to update the posts, it would conclude that the Discourse-served data and the local post data were different, and promptly re-update the posts.  This does not happen on subsequent reconciliations after the first one, because there is logic that keeps (in-memory) whether a post has already been updated or not, and does not attempt to update the posts after that (saving Discourse API requests as the posts won't be checked more than once during the lifecycle of release controller).

We solve this problem by chomping the trailing whitespace out of the post generator that creates the posts for Discourse.

### DIFF
--- a/release-controller/forum.py
+++ b/release-controller/forum.py
@@ -1,7 +1,7 @@
 import difflib
 import logging
 import os
-from typing import cast, Callable, TypedDict, Protocol
+from typing import cast, Callable, TypedDict, Protocol, ParamSpec
 
 from dotenv import load_dotenv
 from pydiscourse import DiscourseClient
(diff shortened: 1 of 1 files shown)


sasa-tomic (approved):
Awesome fix!!! TY!

DFINITYManu:
This worked!

No more updating posts during first reconcile:

image

PR #2023, opened 2026-05-26
fix(release-controller): stop resubmitting election proposals for already-elected versions

pietrodimarco-dfinity:
## Why

The release-controller fired three different NNS election proposals for commit `c7552446...` (141776 -> failed, 141782 -> executed, then duplicates 141879 and 141880, both rejected by governance) because two latent bugs combined badly with a stale `ignored_proposals` entry that nobody removed once 141782 was executed:

1. **`get_election_proposals_by_version` in `public_dashboard.py` and `dre_cli.py`** had a nested loop and a blind `d[version] = proposal` assignment that kept *whichever proposal was iterated last* per version.  The dashboard API returns proposals newest-first, so the *oldest* (the rejected 141776) won.  After `_filtered_proposals_retriever` then stripped 141776, the version looked unproposed on every cycle, and each pod restart submitted yet another one.
2. The proposal-submission phase already fetched the elected (`blessed`) versions to compute `unelect_versions`, but it never used that set to short-circuit the submission, so the governance canister was left to refuse every duplicate on its own.

## What

- Replace the buggy per-version aggregator with **`dre_cli.proposals_by_version`**, a free helper that:
  - keeps the highest-id proposal per version (no reliance on iteration order), and
  - applies `ignored_proposals` filtering at the raw-list level **before** aggregation, so a stale ignored id can't hide a perfectly good non-ignored proposal targeting the same version -- the wrapper falls back to the next-highest-id automatically.
- Refactor `Reconciler._filtered_proposals_retriever` to take a raw-list retriever (`get_past_election_proposals`) and call the new helper, so the filter happens at the right layer.
- Add a `release_commit in blessed` short-circuit to the `proposal submission` phase that warns loudly and skips submission when the target version is already elected -- safety net for any future operator mistake of the same shape.
- Drop the now-stale `141776` entry from `release-index.yaml` and update the README's example to a synthetic id, plus document the new fallback-to-next-highest-id behaviour in the caveats.

## Tests

New tests in `tests/test_reconciler_integration.py`:

- `test_dashboard_get_election_proposals_by_version_keeps_highest_id_for_duplicates` -- guards the dashboard aggregator.
- `test_dre_cli_get_election_proposals_by_version_keeps_highest_id_for_duplicates` -- guards the dre_cli aggregator.
- `test_reconciler_filtered_proposals_retriever_falls_back_to_next_highest_id` -- guards the filter-before-aggregation ordering (this is the user's exact scenario).
- `test_reconciler_skips_submission_when_version_already_in_blessed_set` -- guards the safety net.

Updated tests:

- `test_reconciler_filtered_proposals_retriever_drops_ignored_ids` / `..._is_noop_without_ids` -- migrated to the new raw-list retriever signature.
- `test_reconciler_reconciles_without_error_already_submitted_proposals` -- expected forum-post id updated from `138814` to `138817` to reflect the now-correct highest-id-wins behaviour.

Local runs:

```
bazel test //release-controller:integration_tests //release-controller:medium_tests
# both PASSED
bazel build //release-controller:public_dashboard //release-controller:dre_cli \
            //release-controller:release-controller //release-controller:dryrun \
            //release-controller:integration_tests //release-controller:medium_tests
# mypy clean
```

### DIFF
--- a/release-controller/README.md
+++ b/release-controller/README.md
@@ -224,7 +224,7 @@ next cycle.
 
 ```yaml
 ignored_proposals:
-  - 141776 # GuestOS election proposal rejected; resubmit the version.
+  - 123456 # GuestOS election proposal rejected; resubmit the version.
 releases:
   - rc_name: rc--...
     versions:
@@ -237,9 +237,14 @@ Caveats:
   that is already blessed.  This lever only helps when the prior
   proposal did **not** result in a blessing (typically REJECTED or
   FAILED).
-- Remove the entry from `release-index.yaml` once the replacement
-  proposal has been submitted.  Leaving it in indefinitely silently
-  swallows any future state for the same proposal id.
+- **Remove the entry from `release-index.yaml` once the replacement
+  proposal has been submitted.**  Leaving a stale entry around is
+  harmless for the version it was added for -- the proposal-by-version
+  lookup keeps the largest-id proposal for that version, so a later
+  successful resubmission still takes precedence.  As a safety net, the
+  reconciler also refuses to submit a fresh proposal for a version that
+  is already in the elected (`blessed`) set, logging a clear warning if
+  it sees the situation.
 
 ## Development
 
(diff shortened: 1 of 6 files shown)


NikolaMilosa (approved):

pietrodimarco-dfinity:
Reopening to retrigger CI after self-hosted runner outage cleared.

PR #2039, opened 2026-06-08
fix(cli): download ic-admin from GitHub releases

r-birkner:
## Summary

- `download.dfinity.systems` no longer serves ic-admin binaries; switch to GitHub Releases on `dfinity/ic`
- Add `find_github_release_tag_for_commit` which resolves a commit SHA to a release tag by querying the GitHub `matching-refs` API (searches from current year back to 2024, with pagination)
- Update asset naming to match new conventions: `ic-admin-{arm64,x86_64}-{darwin,linux}.gz`
- Add proper ARM64 support for both Linux and macOS (Apple Silicon)
- Update `FALLBACK_IC_ADMIN_VERSION` to `b95f4a32b41798de115aac9298b51dd1662f1da5` (`release-2026-06-04_04-52-base`), since the old fallback has no GitHub release

## Test plan

- [ ] Run `dre` on a macOS (Apple Silicon) machine and verify ic-admin is downloaded from GitHub
- [ ] Run `dre` on an x86_64 Linux machine and verify ic-admin is downloaded from GitHub
- [ ] Verify the fallback version resolves correctly to a GitHub release tag

🤖 Generated with [Claude Code](https://claude.com/claude-code)


rs/cli/src/store.rs
```diff
@@ -168,15 +254,15 @@ impl Store {
...
-            // Apple Silicon will emulate x86 architecture.
-            format!("https://download.dfinity.systems/ic/{version}/binaries/x86_64-darwin/ic-admin.gz")
-        } else {
-            format!("https://download.dfinity.systems/ic/{version}/binaries/x86_64-linux/ic-admin.gz")
+    async fn download_ic_admin(&self, tag: &str, path: &PathBuf) -> anyhow::Result<()> {
+        let asset_name = match (std::env::consts::OS, std::env::consts::ARCH) {
+            ("macos", _) => "ic-admin-arm64-darwin.gz",
+            (_, "aarch64") => "ic-admin-arm64-linux.gz",
+            _ => "ic-admin-x86_64-linux.gz",
         };
```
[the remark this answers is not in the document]
r-birkner: done

rs/cli/src/store.rs
```diff
@@ -12,6 +12,92 @@ use log::{debug, info, warn};
...
+
+    for year in (oldest_year..=current_year).rev() {
+        let mut page = 1u32;
+        loop {
+            let url = format!("https://api.github.com/repos/dfinity/ic/git/matching-refs/tags/release-{}", year);
+            let mut request = client.get(&url).query(&[("per_page", "100"), ("page", page.to_string().as_str())]);
+            if let Some(token) = &github_token {
+                request = request.bearer_auth(token);
+            }
+            let response = request.send().await?;
```
NikolaMilosa, line 67: I don't know if what I'm going to say is a good or bad idea but: In the `release-index.yaml` we have the map from commits to the `rc--<date>`. Later, we remove `rc--` and replace it with `release-`. Maybe you can reverse map it? That way this pagination won't hit any limits or whatever.

I don't have anything against this approach just mentioning that it exists.

NikolaMilosa (approved):
Thanks!

rs/cli/src/store.rs
r-birkner: The release-index.yaml file is part of the dre repo, but is not shipped with the dre tool itself (e.g., when somebody downloads the latest release), right? In that case, I will leave it as is.

PR #2042, opened 2026-06-08
fix(bazel): repin Cargo.Bazel.lock after #2040 manifest update

pietrodimarco-dfinity:
## What

Regenerate `Cargo.Bazel.lock` (`crate_index_dre`) via `CARGO_BAZEL_REPIN=true bazel sync`.

## Why

PR #2040 ("fix(registry): support NodeRewardType type4.1-4.5") updated `Cargo.toml` and `Cargo.lock` but did not repin the Bazel crate lockfile. This left `crate_index_dre`'s recorded digest stale, so Bazel fails at repo fetch with:

```
Error: Digests do not match ...
The current `lockfile` is out of date for 'crate_index_dre'.
Please re-run bazel using `CARGO_BAZEL_REPIN=true` ...
```

This has been breaking `main` CI ("Build and test", "Required checks") since #2040 merged, and it also blocks the `v0.7.8` release pipeline (its `test` job runs `bazel test` without repin).

## Verification

`bazel query //rs/cli:dre` now resolves cleanly without the digest mismatch.

### DIFF
--- a/Cargo.Bazel.lock
+++ b/Cargo.Bazel.lock
@@ -1,5 +1,5 @@
 {
-  "checksum": "14b9ca83cfd5906d02a0fb365964be4755ab0de3ee8b3191e7daf43c2223475d",
+  "checksum": "9e0131cec0a9bb0bbdc364c7d7492aef9c97103418037447af45cf0188a963fd",
   "crates": {
     "actix-codec 0.5.2": {
       "name": "actix-codec",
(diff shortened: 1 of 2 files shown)


pietrodimarco-dfinity:
Pushed a second commit: the lockfile repin exposed a real compile break that the stale lockfile was masking — #2040 added `ProposalInfo.success_value` and updated the main code path (`proposal.rs`) but missed the test-only `gen_test_proposal` constructor in `slack-notifications`. Now initialized to `None`.

r-birkner (approved):

